PR #61, opened 2021-04-02
Created alert include

c2nelson:
Removed html for alert section from the homepage and created a new alert include file.


hursey013 (approved):
This looks great!

PR #72, opened 2021-04-08
[WIP] Adds USWDS icons to service animals overlay

brentryanjohnson:
[PREVIEW 🦮 ](https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/icons/understanding/service-animals/)
- Adds green check and red "x" icons to service animals overlay
- Adds provisional "info" icon to "other laws..."

### DIFF
--- a/_pages/understanding/service-animals.md
+++ b/_pages/understanding/service-animals.md
@@ -5,7 +5,7 @@ layout: page
 lead: |-
   The ADA explains what businesses and state/local governments must do to make sure that they do not discriminate against a member of the public with a disability who uses a service animal.
 
-  Generally, businesses and non-profits that are open to the public as well as state/local governments must allow service animals to go most places where the public can go. This is true even if they have a "no pets" policy.
+  Generally, businesses and non-profits that are open to the public as well as state/local governments must allow service animals to go most places where the public can go. This is true even if they have a “no pets” policy.
 
 ---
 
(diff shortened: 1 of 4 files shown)


_pages/understanding/service-animals.md
```diff
@@ -5,29 +5,54 @@ layout: page
...
 ## About Service Animals
 
 ### Service animals are:
 
-<ul class="icon-list">
-  <li>Dogs</li>
-  <li>Any breed and any size of dog</li>
-  <li>Trained to perform a task directly related to a person’s disability</li>
-</ul>
+  <ul class="usa-icon-list"><li class="usa-icon-list__item">
```
hursey013, line 16: With kramdown we have the ability to add a class to a markdown formatted list by adding `{: .myclass}` immediately after the block. I haven't tried yet, but what do you think about abstracting as much of this markup as possible into a stylesheet, that way we'd be able to use pure markdown w/ the proper classes applied?

I did something similar a while back here: https://github.com/GSA/plainlanguage.gov/wiki/Add-inline-examples

brentryanjohnson: Yes, using markdown is preferable. I think the class formatting in kramdown requires a parent element (other than `<p>`. Since the list items generate the `ul`, I wasn't sure if the class can be applied with markdown.

brentryanjohnson:
Closing because deprecated by other work.

PR #75, opened 2021-04-13
Update ruby dependencies

hursey013:
This updates Jekyll to the latest version as well as cleans out some unused plugins.


brentryanjohnson (approved):
I was able to pull the branch locally and build it successfully. Docker failed to build, but that's probably not a big deal right now, since DRS isn't using Docker.

PR #77, opened 2021-04-13
Adds alert for user testing

brentryanjohnson:
This PR adds an alert banner to the overlay pages for user testing. Since we're moving from the sandbox to beta for user testing, DRS wants this banner to call out the site as a work in progress.

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -13,6 +13,7 @@ associated with it. For example, an about page. {% endcomment %}
       </div>
       <div class="tablet:grid-col-8 tablet:grid-offset-1">
         <h1>{{page.title}}</h1>
+        {% include alert-proto.html %}
         <div class="crt-landing--separator"></div>
         {% if page.lead %}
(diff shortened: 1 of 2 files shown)


_layouts/page.html
```diff
@@ -13,6 +13,7 @@
       </div>
       <div class="tablet:grid-col-8 tablet:grid-offset-1">
         <h1>{{page.title}}</h1>
+        {% include alert-proto.html %}
```
hursey013, line 16: Could we move this below the `<div class="crt-landing--separator"></div>` for the time being so the heading and decorative accent stay together?

brentryanjohnson: Good catch. I'll push another commit.

hursey013:
Looks like my `h2` styling is a little too greedy, applies the accent line within the alert as well, but I will fix that in a follow up PR.
Screen Shot 2021-04-14 at 11 54 13 AM

hursey013 (approved):

PR #83, opened 2021-04-21
Prevent layout shift on nav link hover

hursey013:
Prevents the small shift in the top navigation links on hover:

Screen Shot 2021-04-21 at 3 11 53 PM


c2nelson (approved):
Thanks.

PR #87, opened 2021-04-21
Make entire card clickable

hursey013:
Makes the entire card area clickable on the landing page and adds subtle hover effect.

Screen Shot 2021-04-21 at 4 44 00 PM


c2nelson (approved):
good change!  thanks.

PR #90, opened 2021-04-22
Add collapsable table of contents to overlay pages

hursey013:
Creates mobile first menu for content pages.

Screen Shot 2021-04-22 at 3 52 45 PM


brentryanjohnson:
[Federalist PREVIEW](https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/bh-mobile-toc/guides/service-animals/)

brentryanjohnson:
Reviewing this, I wonder if we should only include the current page items in the TOC (as opposed to also including the other overlay pages). With only three, it's not too onerous, but eventually we may have 5-10 of these pages (or more). Surfacing all of those in the TOC might be undermine the user benefit of the TOC, especially on mobile.

hursey013:
> Reviewing this, I wonder if we should only include the current page items in the TOC (as opposed to also including the other overlay pages). With only three, it's not too onerous, but eventually we may have 5-10 of these pages (or more). Surfacing all of those in the TOC might be undermine the user benefit of the TOC, especially on mobile.

@brentryanjohnson Totally agree - right now those pages are included because without them there the only way to navigate between the different guides would be to return to the homepage and click on the cards. Ideally I think we could create an ADA Guides landing page that can list out all of the guides, and then when navigating to a particular guide only links relevant to that guide would be displayed in the TOC.

hursey013:
Going to merge this in as is and then revisit this comment above in another issue.

PR #104, opened 2021-05-10
Parking edits and image refactoring

brentryanjohnson:
- Fixes broken link to 2010 ADA standards and clarifies link text
- Converts images to block elements (from grid)


hursey013 (approved):
Looks great. In the future, if we could get higher resolution images that would help on high resolution screens so that the images look crisp.

PR #112, opened 2021-05-17
Mobility devices

c2nelson:
I think the detais/summary re assurances is too close to the red x.  Take a look.

### DIFF
--- a/_assets/css/custom/_page.scss
+++ b/_assets/css/custom/_page.scss
@@ -3,7 +3,7 @@
   padding-bottom: 3rem;
 
   .crt-lead {
-    > p {
+    p, li {
       @include text--heading__medium();
     }
   }
--- a/_config.yml
+++ b/_config.yml
@@ -106,6 +106,8 @@ assets:
   raw_precompile:
     - source: node_modules/uswds/dist/fonts/**/*
       strip: node_modules/uswds/dist/
+    - source: node_modules/uswds/dist/img/sprite.svg
+      strip: node_modules/uswds/dist/
   sources:
     - node_modules/gumshoejs/dist
     - node_modules/uswds/dist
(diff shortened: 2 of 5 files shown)


hursey013 (approved):
This looks great! 🎉

hursey013:
@c2nelson Awesome work, this is perfect! I've opened up another PR (https://github.com/usdoj-crt/beta-ada/pull/115/files) to improve our icon list widget, which should help add some flexibility for bigger blocks of content within the lists.

@brentryanjohnson also suggested pulling the content in the list on the bottom of the page out of the `<details />` block so it's more readable so I added that into my PR above as well. When you have a moment, take a look at https://github.com/usdoj-crt/beta-ada/pull/115/ and see if you have any questions, otherwise once we get that merged in, when can then merge this one in!

colinmurphy01:
@hursey013 I'm seeing this branch merged to main, but still hidden (this is a good thing!)

Curious as to how you're managing these hidden overlays in the main branch?

brentryanjohnson:
@colinmurphy01 It looks like he's including a [front matter key that prevents the page from publishing](https://jekyllrb.com/docs/front-matter/).

`published: false`

hursey013:
@colinmurphy01 https://github.com/usdoj-crt/beta-ada/wiki/Front-matter

brentryanjohnson:
@colinmurphy01 We could publish it in a separate branch with a Federalist preview.

hursey013:
Yes, that is the tradeoff - we avoid merge conflicts and unnecessary complexity by continuously merging into main, however if folks want to see the rendered version of the page they will either need to run the site locally or spin up a temporary branch removing the `published: false` on the page in question.

My hope would be that as we progress this will become a non-issue. The new overlay content can be developed in a new branch, folks can QC/review/approve, and then it will be merged and published on the website. This is a similar workflow that digital.gov uses, where the pull request is opened once the content is already approved, minimizing the time the PR needs to sit there - the final check is basically checking that the styling/rendering of the page looks good, not a thorough content review, that could/should happen before it's moved into code.

PR #114, opened 2021-05-18
Make sticky nav scrollable

hursey013:
Allows left nav to be scrollable when there is not sufficient height to show all of the links (on desktop).


lauraponce (approved):
Works great!

PR #118, opened 2021-05-25
Add alt text to mobile and desktop hero images

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/172


brentryanjohnson:
I tested this with VoiceOver and it works great! The background image description comes before the `h1`, but I think that's okay, since the visual effect is similar.

brentryanjohnson (approved):

PR #120, opened 2021-05-25
Add additional skip link for overlay page layout

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/151

This presents users with two skip nav links - the first will redirect focus to the side navigation containing in-page anchor links, and the second skip nav will allow folks to go directly to the content, bypassing the in-page links.

Screen Shot 2021-05-25 at 4 21 18 PM

### DIFF
--- a/_includes/skipnav.html
+++ b/_includes/skipnav.html
@@ -5,6 +5,6 @@
 {% endif %}
 
 {% if has_side_nav %}
-<a class="usa-skipnav" href="#crt-page--sidenav">Skip to in-page navigation</a>
+<a class="usa-skipnav" href="#crt-page--sidenav">Skip to page navigation</a>
 {% endif %}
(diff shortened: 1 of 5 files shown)


_includes/skipnav.html
```diff
@@ -1 +1,10 @@
-<a class="usa-skipnav" href="#main-content">Skip to main content</a>
+{% if page.layout == 'page' and page.sidenav %}
+  {% assign has_side_nav = true %}
+{% else %}
+  {% assign has_side_nav = false %}
+{% endif %}
+
+{% if has_side_nav %}
+<a class="usa-skipnav" href="#crt-page--sidenav">Skip to in-page navigation</a>
```
brentryanjohnson, line 8: I wonder if "Skip to page navigation" might work? Does that sufficiently imply "in-page"?

brentryanjohnson (approved):

PR #123, opened 2021-06-02
Remove robots meta tag

hursey013:
This tag prevents search engines from indexing pages - we will want to remove this prior to launch.


colinmurphy01:
Per our rollout plan, we are not indexing this site in search engines yet. Unless we've changed the decision on this, CC @kstitch7.

kstitch7:
That is right. We are not indexing the site in search engines yet.

c2nelson (approved):

PR #124, opened 2021-06-02
Remove legal disclaimer

hursey013:
To be removed prior to launch.


c2nelson:
We will remove it once we get final approvals.

PR #130, opened 2021-06-04
Edits title and adds description for homepage

brentryanjohnson:
- Edits homepage title from `Home` to `The Americans with Disabilities Act`
- Adds description field to homepage


hursey013:
@brentryanjohnson one small request - let's move the `title` and `description` out from the landing page and into the main `_config.yml` so it can be used as the fallback sitewide (as well as allowing the homepage to inherit it).

hursey013 (approved):

PR #138, opened 2021-06-08
Update header title

hursey013:
Switches title in header back to `ADA.gov` rather than:

Screen Shot 2021-06-08 at 9 39 29 AM


kstitch7 (approved):
looks good

PR #143, opened 2021-06-25
Ports ADA statute, adds headings, reclasses to USWDS

brentryanjohnson:
- Ports ADA statute
- Adds heading levels for accessibility and in-page navigation
- Reclasses margins to USWDS
- Lighthouse accessibility audit is 🟢  100
- Needs additional work, as indicated by 🔴  linter


brentryanjohnson:
The markdown linter has flagged a valid issue, which reflects ambiguity in heading levels for Title 47. We'll need @kstitch7 to interpret the intended hierarchy so we can structure the heading levels.

brentryanjohnson:
A few more notes:

- This port omits the TOC in favor of side nav (we might want to customize the nav heading level for this one)
- I'm unsure what heading level the two Title 47 headings should be relative to one another
- The heading levels are particular to each subchapter, since some have subparts and/or other subheading levels
- I still need to test with a screen reader
- We should evaluate with the DOJ accessibility specialist

brentryanjohnson:
- Resolved lint issue by removing duplicate subheadings. Need @kstitch7 to review.

_assets/css/custom/_lists.scss
```diff
@@ -0,0 +1,3 @@
+.list-margin-unstyled {
+  list-style-type: none;
```
hursey013, line 2: Not sure if this accomplishes the same thing, but USWDS does provide an unstyled list class with `.usa-list.usa-list--unstyled`.

brentryanjohnson: Good catch @hursey013. I started by refactoring just about everything to unordered and ordered lists, but the `unstyled` uswds class also removes margins. I ended up just using margin classes anyway, so this custom rule can go away.

c2nelson:
@kstitch7 @brentryanjohnson check it out.  should be good to go.

brentryanjohnson:
@kstitch7 Here's a more convenient [preview URL](https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/rj-law-regs-2/law-and-regs/ada/)

kstitch7:
Looks great! Just flagging that the preview has the old version of the header.

hursey013 (approved):
Looks good from an implementation standpoint! 🎉

PR #148, opened 2021-07-01
Add print stylesheet

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/205

Adds a print stylesheet to simplify and increase readability of printed pages:

Screen Shot 2021-07-01 at 3 55 23 PM

### DIFF
--- a/_assets/css/custom/_all.scss
+++ b/_assets/css/custom/_all.scss
@@ -20,3 +20,4 @@
 // Layout specific
 @import "landing";
 @import "page";
+@import "print";
--- a/_assets/css/custom/_print.scss
+++ b/_assets/css/custom/_print.scss
@@ -1,0 +1,20 @@
+@media print {
+  * {
+    color: #000 !important;
+  }
+  #crt-page--content {
+    width: 100% !important;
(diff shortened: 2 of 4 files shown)


brentryanjohnson (approved):
This looks good overall! 

I think my only concerns are (1) the content crowds the top and bottom of the page and might benefit from margin, and (2) selecting all elements via * might be a problem at some point if for example light text is ever used on a dark background (it also lessens the effect of the ✅ and ❌ icons currently). I understand it might be safer to monochrome all the text, but just wanted to flag those! I think it's fine to ship as-is.

hursey013:
@brentryanjohnson updated based on your feedback. The monochrome text is specifically to address those instances when it's white text on a dark background - without the `*` we'd need to add classes/styles to all of those elements to have them displayed differently when printed. With the approach in this PR we can be confident that all text on the page will be clearly displayed.

Also, added some extra margin (good call!), and made sure that the icons kept their respective colors.

Screen Shot 2021-07-07 at 3 45 14 PM

PR #165, opened 2021-07-30
Fix for search pagination links

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/281

### DIFF
--- a/_assets/css/custom/_pagination.scss
+++ b/_assets/css/custom/_pagination.scss
@@ -3,3 +3,7 @@
   opacity: 1 !important;
   pointer-events: none;
 }
+
+.usa-pagination[hidden] {
+  display: none !important;
+}
--- a/_pages/search.md
+++ b/_pages/search.md
@@ -73,18 +73,22 @@ sidenav: false
       var nextLink = document.querySelector(".usa-pagination__next-page");
       var currentOffset = urlParams.get("offset");
 
-      if (currentOffset > 0) {
-        urlParams.set("offset", currentOffset - {{site.searchgov.limit}});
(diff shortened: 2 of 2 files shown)


kmori (approved):
Looks good! Confirming that the "Next" and "Previous" are no longer on the page when there are no results
image

kmori:
Should we remove the "Previous" and "Next" when there's only one page worth of results?

hursey013:
> Should we remove the "Previous" and "Next" when there's only one page worth of results?

@kmori We could, although I believe with just one page of results both the next and previous links would appear but in a disabled state (unclickable). If we think it makes more sense for them to not appear at all with less than 20 results, let me know and I can update.

kmori:
> > Should we remove the "Previous" and "Next" when there's only one page worth of results?
> 
> @kmori We could, although I believe with just one page of results both the next and previous links would appear but in a disabled state (unclickable). If we think it makes more sense for them to not appear at all with less than 20 results, let me know and I can update.

@lauraponce- any thoughts from a UX perspective of if we should keep the "Previous" / "Next" in a disabled state when there's only one page of search results or remove it?

lauraponce:
I think it would be better to remove them rather than keep the links there and disabled since it might not seem clear to people why the links are disabled.

Is it possible to add the `pages 1/1` at the bottom near the buttons to show how many pages of results there are? (This also is out of scope from this ticket so it shouldn't hold up this one from closing and could go in another one if needed.)

kmori:
thanks @lauraponce ! @hursey013 - can we remove the links when there's only one page of results. I'll also make a new ticket for the recommendation provided above re number of pages

PR #166, opened 2021-07-30
Add search results sort text

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/268

Adds a message that the results are sorted by relevancy.

Screen Shot 2021-07-30 at 11 26 54 AM


austinhernandez (approved):
Looks good!

PR #167, opened 2021-07-30
Adds margin levels for statute indents

brentryanjohnson:
- Adds custom margin levels to extend USWDS pattern
- Applies more aggressive indents for use on ADA statute
- Crosswalks margin styles for statute indent

[Preview 🔍 ](https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/rj-law-margins/law-and-regs/ada/)

### DIFF
--- a/_assets/css/custom/_utilities.scss
+++ b/_assets/css/custom/_utilities.scss
@@ -22,17 +22,3 @@
 .grid-gap-xl {
   @include grid-gap(6);
 }
-
-// Custom margin additions
-
-.margin-left-12 {
-  margin-left: 6.5rem;
-}
-
-.margin-left-16 {
-  margin-left: 8rem;
-}
-
-.margin-left-20 {
-  margin-left: 10rem;
-}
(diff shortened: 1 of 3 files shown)


kstitch7 (approved):

_assets/css/custom/_utilities.scss
```diff
@@ -22,3 +22,17 @@
 .grid-gap-xl {
   @include grid-gap(6);
 }
+
+// Custom margin additions
```
hursey013, line 26: I think this is perfectly fine, so feel free to take or leave this suggestion, but another way to accomplish this would be to add these values to `_uswds-theme-utilities.scss` like:

```
$margin-manual-values: (12: 6.5rem, 16: 8rem, 20: 10rem);
```
Main benefit with this approach would be for consistency, making those new steps available for all variations of margin.

brentryanjohnson:
Got it, @hursey013. Great suggestion! I added a new commit ☝️ if you have a chance to review.

hursey013:
Looks good!

PR #169, opened 2021-08-06
Update default card image

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/282

Updates image so it is properly aligned:

Screen Shot 2021-08-06 at 12 13 13 PM


kmori (approved):
Screenshot in the ticket looks good. I couldn't verify in the preview because we have images for all the cards now

PR #177, opened 2021-08-17
Changes 'ADA Topics' to 'Topics'

brentryanjohnson:
Changes nav text, page heading, and homepage section heading to `Topics` from `ADA Topics`.

The need for concision in the nav list made me think that `ADA Topics` is increasingly redundant. We developed that term as a working label for the overlay content, but now that we have a wordmark and more surrounding content, it seems _possibly_ unnecessary. I'm certainly willing to be persuaded otherwise, but I thought I would send it up to a pull request in the meantime to generate a preview.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -45,7 +45,7 @@ google_site_verification: 34Ks265Uibpdda984RS0SxVABVZCdlVrdSp61pIdZQ8
 primary_navigation:
   - name: Home
     url: "/"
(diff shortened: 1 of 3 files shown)


kstitch7 (approved):

lauraponce (approved):
This seems like a good idea to me 👍  It works well in the nav and seems clear as the h1 on the page

PR #179, opened 2021-08-20
Adds missing punctuation to home independent clause

brentryanjohnson:
Adds missing punctuation to `We‘ll add more soon` on homepage.

### DIFF
--- a/_pages/index.md
+++ b/_pages/index.md
@@ -55,8 +55,6 @@ report:
   heading: How to Report a Disability Rights Violation
   lead: |-
     If you believe that you or someone else experienced unlawful discrimination, you can report a disability rights violation.
-  content: |-
-    <strong>Pick a topic to find out how the ADA works.</strong> We’ll add more soon.
   steps:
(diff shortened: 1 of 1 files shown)


_pages/index.md
```diff
@@ -56,7 +56,7 @@ report:
   lead: |-
     If you believe that you or someone else experienced unlawful discrimination, you can report a disability rights violation.
   content: |-
-    <strong>Pick a topic to find out how the ADA works.</strong> We’ll add more soon
+    <strong>Pick a topic to find out how the ADA works.</strong> We’ll add more soon.
   steps:
```
hursey013, line 60: Oops, this must have been a copy/paste error on my part - this content block can be removed since it is not actually being used by the reporting section.

hursey013 (changes requested):

_pages/index.md
brentryanjohnson: I was wondering about that. I'll push another commit.

hursey013 (approved):

PR #180, opened 2021-08-24
Use collections for topics

hursey013:
Fixes https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/305

This organizes the topics content into Jekyll's organizational structure called `collections`. It allows us to access and control data by specifying what type of content it is rather than using the url to determine it. So the variable `site.topics` would retrieve all ADA topic pages, making it easier to sort and display.

One benefit of this approach is the built in ordering that's included with collections, so now rather than needing to specify weights, you can just define the order you'd like:

```
    order:
      - intro-to-ada.md
      - service-animals.md
      - parking.md
```

This can be controlled in the `_config.yml` file.


brentryanjohnson (changes requested):
@hursey013 It looks like the paths aren't generating properly. The topics links are broken.

brentryanjohnson (approved):

PR #199, opened 2021-10-12
Replace form 3

billyhunt:
I updated the config to exclude anything we put in the download directory.  We should be able to link to any pdf's or whatever we put in that folder now, without tripping up Pa11y.


billyhunt:
updated with new version of pr.

PR #210, opened 2021-12-17
Chrome Sidenav/Main Content Bug Fix

JackRyan1989:
Added dummy divs and and javascript

### DIFF
--- a/_assets/js/skip-content-links.js
+++ b/_assets/js/skip-content-links.js
@@ -1,0 +1,17 @@
+//= require ./scroll-into-view.js
+// Grab our links:
+var skipSideNav = document.getElementById('crt-skipnav-sidenav');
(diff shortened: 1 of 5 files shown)


_assets/js/skip-content-links.js
```diff
@@ -0,0 +1,17 @@
+//= require ./scroll-into-view.js
```
billyhunt, line 1: I like that you nested this.  We can start breaking up the js bundles perhaps for performance down the road.

billyhunt (approved):
Looks good to me.  Elegant solution!

JackRyan1989:
Federalist Link tested using BorwserStack on following browsers:
Edge (multiple versions)
Safari (13.1)
Opera 
Chrome (multiple versions)

PR #211, opened 2021-12-17
Bundler updated to version 2.2.33 - WIP do not review

JackRyan1989:
Seeing if updating Bundler breaks Federalist


JackRyan1989:
Noting here: 
bundler v 2.2.33 breaks federalist build process

JackRyan1989:
Noting here: webrick v 1.17.3 does not break federalist

JackRyan1989:
Confirmed web 1.17 does not crash Federalist.

JackRyan1989:
Babel passes.

PR #212, opened 2021-12-21
Pa11y update

JackRyan1989:
Changes:
1. Added a `pa11y-skip` class for a11y issues that are manually addressed.
2. So far this class has been added to the following:
a. The homepage H1 and P tags. pa11y flags them because they sit over top of an image. Manual inspection of the text over a white background reveals a AAA rating (WCAG2.1). A minor PR was approved a few weeks ago to add some additional spacing on tablet sized devices so that the text minimally overlaps the part of the image with people in it.
b. The highlighted nav item. Pa11y can't correctly assess pseudo elements, in this case the underline on the current nav location. The line color is the same as the text, and also scores a AAA in contrast (WCAG2.1).
c. Accordion plugin - The button uses a background image for the plus and minus icons. Pa11y can't accurately asses the contrast when a background image is used, so the `pa11y-skip` class is added to the returned HTML in the button element. Manually confirmed to have a greater than AA contrast.
d. Details plugin - Same issue as the accordion, contrast for the text was confirmed manually to be over AA contrast.


JackRyan1989:
Where we're at:
1. Updated pa11y-ci
2. Updated super linter to newest version
3. Added stylelint standard scss to repo. 
4. We now have a ton of linting errors. 
5. Still not sure if this will fix the the pa11y timeout errors.

JackRyan1989:
Updates:

1. Added custom class called ".pa11y-skip" that tells pa11y not to check the element that the class is added to. Currently skipping accordion and details elements. I manually confirmed that the text and backgrounds meet WCAG AAA standards for color contrast with ratios going from ~12:1 to 13.75:1. Note the class name has to be added in the ruby files in the _plugins folder.
2. Added ".usa-current.usa-nav__link" to the hideElements array so that the currently selected nav element does not trigger a test failure. Manually confirmed that the contrast is WCAG AAA with a ratio of 13.75:1.

billyhunt:
Heyo!  Will you add some details to the PR?  Overview, steps to test, what should I be looking for, etc?

JackRyan1989:
This needs to be cleaned up. I'm going to make a new pull request.

PR #214, opened 2021-12-28
Added babel transpiler v 0.7

JackRyan1989:
[Babel Transpiler](https://rubygems.org/gems/babel-transpiler/versions/0.7.0) will allow developers to write safer, modern javascript in our code base while maintaining backwards compatibility with older browsers.

This change is in the Gemfile, which means that users will need to run `bundler install` after pulling in this change to make sure the software is installed locally. 

Babel Transpiler is run at build-time and converts all Javascript files that we denote with an .es6 extension to backwards compatible versions of Javascript.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ ruby '>= 2.6.6'
 
 gem "jekyll", "~> 4.0"
 gem "autoprefixer-rails"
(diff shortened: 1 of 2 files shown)


JackRyan1989:
So after some more thought, we're not going to include this package. It's old, unmaintained and locked to an old version of babel. I'm going to close this ticket.

PR #215, opened 2022-01-04
Super linter update

JackRyan1989:
Updated to version 4.

Change in: /beta-ada/.github/workflows/linter.yml
Updated line 64 to `uses: github/super-linter@v4.0.2` from `uses: github/super-linter@v3`

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -58,7 +58,7 @@ jobs:
       # Run Linter against code base #
       ################################
       - name: Lint Code Base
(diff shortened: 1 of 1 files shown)


JackRyan1989:
One thing to consider is that we are now locked to a specific version of Super Linter. I'll monitor changes and see when we might want to upgrade. When we make that decision, we'll need to upgrade StyleLinter as well.

c2nelson:
thank you, jack, for your thorough and expeditious work on this.

c2nelson (approved):

PR #223, opened 2022-01-11
Fix Last page of results no results bug

JackRyan1989:
Three main changes: Removed lingering arrow funcs, capped results at 1000 and made sure that the search pages end at the right spot.

Closes: https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/372


billyhunt (approved):
Looks Good to me!

PR #229, opened 2022-01-25
Chrome sidenav issue demonstration for USWDS

JackRyan1989:
This is to demonstrate the Chrome bug that happens when using the Skip Side Nav and Skip to Main Content buttons. 

To replicate the issue:

1. Open Chrome on Desktop
2. Navigate to the url of the Federalist build preview
3. Navigate to a page with a side nav (/topics/intro-to-ada for example)
4. Press TAB to expose the skip side nav button
5. Select the skip side nav button
6. The side nav should disappear from the page
7. Repeat for the skip main content button - it should also disappear

Hypothesis:
1. Setting the outline to 0 is at fault - in fact it seems that setting any inline style on the element removes it from the page.
2. The element is still in the DOM.
2. This occurs in line 20 of /uswds/src/js/components/skipnav.js : `target.style.outline = "0";`
3. I am unsure as to why this is happening.


JackRyan1989:
Mentioned in USWDS Issue 1610: https://github.com/atbcb/usab-uswds/issues/1610

bruce-usab:
@JackRyan1989 USAB has a similar difficulty with some sticky navigation components in mobile view.  Ping to @RSD-accessboard since he has worked on the most and might be able to offer some insight.

The pages respond with the second or third down-arrow/swipe but it is not right.  We have not come up with a solution which seems less harmful than the problem (e.g., browser sniffing).

JackRyan1989:
@bruce-usab Interesting! Can you point me to a place where I can see that issue in action?

JackRyan1989:
Issue was due to UBlock Origin extension enabled in Chrome. Disabling the extension resolves the issue. Will close issue in USWDS as well.

Screen Shot 2022-01-28 at 11 27 33 AM

PR #230, opened 2022-01-28
Reverting chrome bug fix.

JackRyan1989:
Changes:
1. Deleted _assets/js/scroll-into-view.js
2. Deleted _assets/js/skip-content-links.js
3. Removed additional HTML attributes (id and data-scroll-to) from skipnav.html
4. Removed dummy divs from page.html


billyhunt (approved):
Looks good to me.  Nice work investigating the root cause!

PR #231, opened 2022-01-31
Pa11y update to 3.0.1

JackRyan1989:
Changes:
1. Added a `pa11y-skip` class for a11y issues that are manually addressed.
2. So far this class has been added to the following:
a. The homepage H1 and P tags. pa11y flags them because they sit over top of an image. Manual inspection of the text over a white background reveals a AAA rating (WCAG2.1). A minor PR was approved a few weeks ago to add some additional spacing on tablet sized devices so that the text minimally overlaps the part of the image with people in it.
b. The highlighted nav item. Pa11y can't correctly assess pseudo elements, in this case the underline on the current nav location. The line color is the same as the text, and also scores a AAA in contrast (WCAG2.1).
c. Accordion plugin - The button uses a background image for the plus and minus icons. Pa11y can't accurately asses the contrast when a background image is used, so the `pa11y-skip` class is added to the returned HTML in the button element. Manually confirmed to have a greater than AA contrast.
d. Details plugin - Same issue as the accordion, contrast for the text was confirmed manually to be over AA contrast.  
3. Edited `.pa11y-ci` config file so that concurrency is set to 1 as recommended and that the new `pa11y-skip` class is added to the list of items to ignore.

### DIFF
--- a/.pa11yci
+++ b/.pa11yci
@@ -1,11 +1,11 @@
 {
   "defaults": {
-    "concurrency": 4,
+    "concurrency": 1,
     "chromeLaunchConfig": {
        "ignoreHTTPSErrors": true,
        "args": ["--disable-dev-shm-usage"]
      },
-    "hideElements": "#fba-button",
+    "hideElements": ["#fba-button", "pa11y-skip"],
     "standard": "WCAG2AA",
     "runners": ["axe"],
     "timeout": 120000
--- a/.pa11yci
+++ b/.pa11yci
@@ -5,7 +5,7 @@
        "ignoreHTTPSErrors": true,
        "args": ["--disable-dev-shm-usage"]
      },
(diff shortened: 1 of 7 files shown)


.pa11yci
```diff
@@ -1,11 +1,11 @@
 {
   "defaults": {
-    "concurrency": 4,
+    "concurrency": 1,
     "chromeLaunchConfig": {
        "ignoreHTTPSErrors": true,
        "args": ["--disable-dev-shm-usage"]
      },
-    "hideElements": "#fba-button",
+    "hideElements": ["#fba-button", ".pa11y-skip"],
```
billyhunt, line 8: Smart implementation.

billyhunt (approved):
Looks good to me!

PR #235, opened 2022-02-07
Added production jekyll environment to build script.

JackRyan1989:
Changes:
1. Added `JEKYLL_ENV=production` to our build script in `package.json`.
2. I did not add a value of 'development' anywhere because [Jekyll assumes that value for the environment if none is specified](https://jekyllrb.com/docs/configuration/environments/)
3. The goal here is to prevent Google Analytics from running when on localhost, so in `scripts.html` I've added an and clause to the if statement that injects the GA script: `if site.ga.ua and jekyll.environment=="production"`

### DIFF
--- a/_includes/scripts.html
+++ b/_includes/scripts.html
@@ -6,7 +6,7 @@
 ></script>
 {% endif %}
 
-{% if site.ga.ua %}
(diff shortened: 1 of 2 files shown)


billyhunt (approved):
Looks good to me!

PR #245, opened 2022-02-28
Webpack Babel Integration

JackRyan1989:
Changes:
1. Added browserslistrc. Settings are taken from the CRT Portal project with one change, IE11 has been removed from the list. 
2. Added /_assets/js/dist to the .gitignore file
3. Added a .prettierignore file. Added the /_assets/js/dist/*.js files.
4. Added a .prettierrc file. Settings are taken directly from the CRT Portal file.
5. Added babel.config.js - these settings are pretty bog-standard. We use babelPresetEnv so we can leverage the power of the browserslist to only include the polyfills as needed. CoreJS is required for the polyfills, and I've specified the version we should use so that Babel doesn't make any assumptions.
6. docs/webpack-babel-decision-record.md: I added an architectural decision record so that we can track the reasons why I'm integrating Webpack and Babel. This should be fleshed out and rewritten.
7. package.json - **Added new scripts.** `build:js` just runs webpack to build out the JS. `clean:js` uses Prettier to reformat all JS files. `lint:js` runs Prettier to see if there are issues with the formatting in any of the JS files. `start:js` runs the webpack dev server and watches & rebuilds if the JS changes. **Changed scripts.** `build` now runs the `build:js` script first. `start` now runs the `start:js` script so the js is rebuilt when it's changed. This then triggers a rebuild of the Jekyll site.
8. package.json - New dependencies - `core-js` Needed for Babel Polyfills. `babel/core` and `babel/preset-env` to run transpilation. `babel-loader` so that `webpack` can pull in and run Babel. `prettier` for JS linting. `webpack` and `webpack-cli` to run bundling and code minification.
9. webpack.config.js - Mode (production or development) is specified by CLI arguments. Currently we are just trying to bundle the pagination scripts, so there is a single entry point for that set of scripts. Additionally, I'm specifying a single output file with the name hardcoded. The rules specify that we're using Babel and that we should exclude our Node Modules folder.


billyhunt (approved):
Worked though this in pairing.  Looks good to me! 

Nice work.

PR #246, opened 2022-03-01
JavaScript Refactor to use ES6 modules

JackRyan1989:
Changes:
All JS files: 
1. Removed the Sprockets convention of requiring in modules using `//require "module"`
2. Replaced with ES modules. Note that in some scripts the revealing module pattern was used along with IIFEs, these have been removed in favor of `import` and `export`.
3. A few `vars` were replaced with `let` and `const`. This is not done universally as I decided to focus on the `import/export` statements for this PR. 
4. Note that all functions are the default export unless there are multiple items being exported from the module.
5. Change to `scripts.html` - Made sure the pagination script is imported from the `/dist` folder.

To Test:
Basically we just need to make sure that all of the JS still functions as before.
1. Check that the 'skip to main content' and 'skip to page navigation' buttons still work. Press 'tab' when first arriving on a page to get those buttons to appear, then press 'enter'.
2. Confirm that the accordions open properly. Confirm that they close other accordions that are already open. The accordions look like this: Screen Shot 2022-03-02 at 9 48 27 AM
3. Confirm that the top 'How you can tell you're on a .gov website' banner opens.
4. Confirm the print button on the Topics pages triggers page printing
5. Confirm that the search feature works
6. Confirm that the search pagination works - should display a total number of results, with 20 results per page, and a list of buttons corresponding to the total number of pages below.
7. Confirm that the urls in the search results wrap on small screen sizes.


JackRyan1989:
Replace vars with let and const

_assets/js/templates/pagination/paginationButtonTemplate.js
```diff
@@ -2,9 +2,9 @@
 import {getOffsetParam, setOffsetParam} from "../../utils/offsetUtils";
 
 export default function paginationButton(offsetVal, index, resultsArr) {
-  var currentOffset = getOffsetParam();
-  var ariaLabel = "";
-  var ariaCurrent = "false";
+  const currentOffset = getOffsetParam();
+  let ariaLabel = "";
+  let ariaCurrent = "false";
```
billyhunt, line 7: Not part of the PR, but should these be boolean?

billyhunt (approved):
Maybe change that one let, but looks good to me!

_assets/js/templates/pagination/pageNumber.js
```diff
@@ -1,19 +1,19 @@
...
       currentPage = resArray.indexOf(currentOffset) + 1;
     } else {
       currentPage = resArray.indexOf(currentOffset) + 2;
     }
-    var lastPage = resArray.length;
-    var templateContent = `${currentPage} of ${lastPage}`;
-    var para = document.createElement("p");
-    var newContent = document.createTextNode(templateContent);
+    const lastPage = resArray.length;
+    let templateContent = `${currentPage} of ${lastPage}`;
```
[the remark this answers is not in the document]
JackRyan1989: Thanks, yep!

_assets/js/templates/pagination/paginationButtonTemplate.js
JackRyan1989: No because they need to go into the template as HTML attributes, so those are always strings. Or usually. At least in this case.

PR #250, opened 2022-03-04
Changed webtotalresults to a mutable variable from const

JackRyan1989:
Bug:
1. When searching for the term 'home', webtotalresults would need to be mutable (defined using 'let') because the value of results is capped at 1000. However it was defined as a constant, so the variable could not be reassigned, and the search feature would break.

Changes:
1. Webtotalresults is defined using let in 'rendersearchpage.js'

Test: 
1. Search for the term 'Home' and it should display results.


c2nelson (approved):
thank you, jack.

PR #251, opened 2022-03-04
News widget 390 

JackRyan1989:
Rationale:
1. We need a place to show off some of the awesome and important things the ADA is doing. 

Changes:
1. New files: `news.html` and `news-item.html`
2. `news.html` is a wrapper that contains all of the logic for displaying the News Component. It first searches all pages for available news items (denoted in the frontmatter by `news-item: true` and `publication-date: 2022-03-06 00:00:00` variables). Puts those items into a list and provides that list to the `news-item.html` component. Unless the list of news-items is 0 items long, we display the news items.
3. `news-item.html` is based off of the [Collection component from USWDS](https://designsystem.digital.gov/components/collection/)
4. A new custom SCSS file was created for custom styling of the Collections component. Currently the max width is fixed to not exceed the length of the subheader.
5. The `news.html` component was included in the `index.md` file.


c2nelson (approved):
great work!

PR #252, opened 2022-03-04
Translate Body Copy only

JackRyan1989:
Rationale:
1. To bring our translated page into alignment with suggestions from the LEP (Low English Proficiency) website recommendations, we have opted to only show the body content of the page in Spanish.

Changes:
1. The 'lang' variable is set to 'en' in the following components: `banner.html`, `banner-beta.html`, `footer.html`, `header.html`, `logo.html`, `notice.html`, `phone.html`, `redirect-modal.html`, `touchpoints.html`
2. This change ensures that the components will only pull in data from the english i18n file, and also only the english names and urls for navigation.
3. `header.html` has another change in it for setting the active class on the current navigation link. Since we only have one page at the moment we are just saying "If the basedir is "es" and you're on the presente-una-queja page then make this the active item.
4. Presente-una-queja.md - added (en ingles) to the link text

### DIFF
--- a/_includes/banner-beta.html
+++ b/_includes/banner-beta.html
@@ -1,4 +1,4 @@
-{% assign lang = page.lang %}
+{% assign lang = "en" %}
 {% if lang==nil %}{% assign lang="en" %}{% endif %}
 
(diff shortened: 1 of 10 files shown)


vwilliamsn (approved):
Thanks @JackRyan1989! Looks good!

PR #260, opened 2022-03-21
Update Pa11y to allow for mobile viewport testing.

JackRyan1989:
Rationale:
1. I'd like to increase our testing coverage to include mobile screen sizes. As of now, we are only running our Pa11y tests on devices at 1280 x 1024 px, which is the default screen size unless otherwise specified. 

Changes:
1. Added new job to Github actions called `Run Accessibility Tests -- Mobile` - this calls our test script
2. Renamed the .pa11y-ci to .pa11y-ci-desktop configuration
3. Added the mobile configuration file .pa11y-ci-mobile. This configuration specifies the viewport size and also the flags the device as a mobile device. 
4.  Created two scripts in the package.json that refer to each pa11y configuration file. Github actions points to each of these scripts and runs them.
5. Minor change to the webpack config to allow for multiple entry points.
6. Added the `pa11y-skip` class to a number of elements that we causing false alarms during a11y scanning.

### DIFF
--- a/_pages/file-a-complaint.md
+++ b/_pages/file-a-complaint.md
@@ -40,7 +40,7 @@ There are three options for filing an ADA complaint:
 1. ### Online <br/>
 <span>File a complaint by submitting a report on the Department of Justice's [Civil Rights Division website](https://civilrights.justice.gov/report/).</span>
 2. ### Mail <br/>
-<span>Fill out and send the paper [ADA Complaint Form](https://www.ada.gov/CRT-ReportPDF-Sep2021.pdf) or a letter containing the same information, to:</span>
+<span>Fill out and send the paper [ADA Complaint Form](https://www.ada.gov/CRT-ReportPDF-Sep2021.pdf){: .pa11y-skip} or a letter containing the same information, to:</span>
 >U.S. Department of Justice<br/>
 Civil Rights Division<br/>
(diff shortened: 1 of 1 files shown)


JackRyan1989:
Terri needs to be consulted about this issue. Resolution of this pull request is going to go till next Thursday.

JackRyan1989:
Looks like toggle menu JS worked. I am now investigating with USWDS why this approach was not taken before.

JackRyan1989:
Ok, update time. After talking to USWDS setting aria-hidden to true should be sufficient to remove an item and it's children elements from the accessibility tree. I confirmed that this is the case on our website. So there is no issue with USWDS. After some more digging it appears that the issue is with Axe-Core (which is the testing software Pa11y is built off of): https://github.com/dequelabs/axe-core/issues/3430
and https://github.com/act-rules/act-rules.github.io/issues/1824

I've already confirmed that the correct items in the node are ignored in the Accessibility tree, and that the menu is indeed functioning as we want using iOS voice over. 

I think the additional JavaScript I have written is not necessary. 

Next steps:
1. Remove new JS
2. Revert webpack.config.js to prior state
3. Remove open menu action from the mobile pa11y test.

I think we can reasonably push this change live. I will confirm with Terri on Thursday though.

c2nelson (approved):
this is great.  thanks, jack!

PR #262, opened 2022-03-23
Alt tag fix for close button.

JackRyan1989:
Change: Removed dang pesky double quotes from language selector in `header.html`.


c2nelson (approved):
thanks!

PR #268, opened 2022-03-31
Mobile menu a11y fix

JackRyan1989:
Issue:
When using accessibility tools on a mobile device, like an iPhone or Android phone, the navigation items in the menu are not selectable.

Cause: Improper markup/JS combo. The JS expects a header element to wrap the navigation div, and instead we have a div wrapping the header element. 

Change: in `header.html` we wrapped the component markup in a header element. 

Test: 
1. Run preview page in iOS with voice over turned on.
2. Run preview page in android with voice over equivalent turned on.


JackRyan1989:
@c2nelson @vwilliamsn @chinelo-ikejimba Works in Chrome browser on desktop, on Android with Accessibility features turned on, and on iOS with Voice Over.

vwilliamsn (approved):
Yay!!! Great job @JackRyan1989!

PR #270, opened 2022-04-01
Multi select accordions

JackRyan1989:
Changes:

1.  The accordion plugin `/_plugins/accordion.rb` can now check if there is an option passed to it in the markdown called 'multiselect'. 
2.  If that option is passed, an aria attribute called  `aria-multiselectable` is added to the containing div element. The US Web Design then recognizes this accordion as multiselectable, and now multiple drop downs can be open at once. 

Testing:

1.  I have added the argument of multiselect to the accordion on the Intro to the ADA page. 
2.  Open the page in the browser/device of choice and confirm that multiple drop downs can be open at once. 
3.  Note: this argument can be passed only to the individual accordion element. So if there are many accordion elements on the page, they will each need this argument passed to them.

NOTE: Do not merge this branch. I need to undo the change to the markdown on the test page prior to merging.


eileen-nava (approved):
Looks good!  💪

eileen-nava (approved):
This is my stamp of approval. 💯

PR #293, opened 2022-04-28
Download pdf button 447

JackRyan1989:
Rationale:
We want to provide a quick way for users to download available pdfs as pdfs without needing to print or save them. In this PR we are creating a new includes component that takes an input pdf name and title and generates an anchor link in a div wrapper. The name is necessary to associate the link with the pdf document to download, and the title attribute is necessary to display the name/title of the pdf the person is downloading. 
Note pdfs need to be stored within the /pdfs/ folder in the main beta-ada directory. That directory path is hard coded into the component. 

Changes:
1. /_includes/download-pdf.html - the component
2. /_pdfs/ - the location where new pdfs need to be stored - added .gitkeep file
3. For demonstration/testing purposes the component is being called into the /web-guidance.md/ file. THIS CONTENT HAS BEEN REMOVED
4. _config.yml - the assets raw-precompile step now includes the _pdfs directory so that pdfs are pulled into the /assets folder at build time.


chinelo-ikejimba (approved):
Yep @JackRyan1989! The changes show no Download button.

PR #297, opened 2022-05-11
Covid vax fact sht

c2nelson:


vwilliamsn (approved):
Great work @c2nelson!

PR #305, opened 2022-06-08
Footnotes Plugin

JackRyan1989:
This pull request adds a new footnotes plugin so we can avoid using kramdown for automatic footnote creation and manually creating footnotes in HTML. 

To do: Review with @c2nelson 

Passes a11y testing with Waves and axe Devtools.


JackRyan1989:
@c2nelson This is fine to merge in.

c2nelson (approved):
thx, jack.

PR #311, opened 2022-06-15
Small change to fix routing with sa ta pages.

JackRyan1989:
Super small fix - removed permalinks from ADA Requirements: Service Animals and Frequently Asked Questions about Service Animals.

This fixes URL issues. Links and everything still work, the URLs on the pages are not accurate though.


c2nelson (approved):
thanks.

PR #315, opened 2022-06-21
updated cases info with link to new cases location

c2nelson:
updated cases info with link to new (present) cases location on DRS' CRT page


vwilliamsn (approved):
Looks good, thanks for your work on this @c2nelson!

PR #320, opened 2022-06-28
Added script to meta tags and added fallback iframe as well.

JackRyan1989:
Changes:

1. Added Google Tag Manager JavaScript to Meta tag.
2. Added fall back iFrame in noscript tag to beginning of body in HTML.

@c2nelson confirm that the the site behaves as normal. There shouldn't be any changes.

### DIFF
--- a/_includes/meta.html
+++ b/_includes/meta.html
@@ -26,4 +26,18 @@
   <!-- CSS
     ================================================== -->
   {% asset styles.css %}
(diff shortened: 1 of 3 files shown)


c2nelson:
@JackRyan1989 I don't see any.  Seems ok.

c2nelson (approved):
Approved.

PR #321, opened 2022-06-28
Removing small console log error from homepage.

JackRyan1989:
The Expand All Accordions Javascript was looking for an element on the page that did not exist on the Home Page, and was leading to the script failing on load. This small change checks that and expand all button is in place in the DOM before firing the DOM Mutation Observer, which removes the error.

To Test:

1. Pull down the branch locally
2. Run npm run test:ui
3. All the tests should pass
4. Navigate to the /resources/service-animals-faqs/ page and confirm that the Open All Sections button does indeed open all of the accordions below it. It should also close them all as well. 
5. Confirm that the text changes to Close All Sections when one is opened.
6. Perform steps 4 & 5 on the Federalist preview link as well. 
7. Ignore the changed Snapshot files. I had to update them to account for the bread crumbs.


c2nelson (approved):
Thx!

JackRyan1989:
@billyhunt confirm that no console log error appears on any of the pages. Here's the specific one we want to avoid:
Screen Shot 2022-06-28 at 4 29 24 PM

billyhunt (approved):
Looks good to me!

PR #322, opened 2022-06-28
added expand all sections accordion

c2nelson:

PR #324, opened 2022-06-29
Subnav Plugin Refactor

JackRyan1989:
Refactored the Subnav creation plugin.

Functionality has changed so that we loop over all HTML elements on the page, and grab the H2 and Paragraph tags.

If it's a paragraph tag we do a further check to see if there is additional text of "expand-heading=" in it. If there is, we add that item to the left side navigation element. 

Note this pull request pulls in pages from another open PR: https://github.com/usdoj-crt/beta-ada/pull/322

This PR should be merged in immediately after that one. 

To Test:
1. Confirm that the side navigation elements have the same ordering as the H2's in the document. Do in Federalist Preview @c2nelson .


c2nelson (approved):
Approved.

PR #326, opened 2022-06-30
Updating ARIA expanded labels for Mobile Menu button

JackRyan1989:
So, I think the Aria Expanded values don't ever need to change. The one button is only visible when the menu is expanded, and the other button is only visible when the menu is closed, so the aria expanded state of each button can remain static.

The states are:
1. Menu closed: the menu open button has an aria-expanded value of 'false'
2. Menu open: the menu close button has an aria-expanded value of 'true'

Tests:
1. Confirm that the aria values are set on each button and read using a screen reader.
2. The close menu button now reads 'close menu' instead of just close.


JackRyan1989:
@c2nelson we'll need to run this by Terri too prior to approving.

JackRyan1989:
@c2nelson Terri has approved this option.

c2nelson:
Yay Jack!

c2nelson (approved):
Thank you!

PR #327, opened 2022-06-30
[WIP] Testing to see if Federalist can build the site properly using Node

JackRyan1989:
Just a test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "federalist-uswds-jekyll",
   "version": "1.3.0",
   "scripts": {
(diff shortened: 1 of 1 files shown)


billyhunt (approved):
Looks good!  Just curious, why did we need to move the dev dependencies into dependencies?

JackRyan1989:
@billyhunt Federalist only installs items located in the dependencies list. I had webpack etc. in the devdeps list and it wasn't installing them. The npm script was erroring out with 'webpack not found' message, so that was a give away that it wasn't installing them.

PR #329, opened 2022-07-12
Fixing the .gitignore file so we stop getting JS merge conflicts

JackRyan1989:
Following a procedure supplied by the one and only @billyhunt.

1. Delete the line from gitignore. (meaning _assets/js/dist)
2. Delete the actual compiled assets. (meaning _assets/js/dist)
3. Push to repo.
4. Add line back to gitignore (meaning _assets/js/dist)
5. Compile assets.
6. Push to repo (this will only push the gitignore file)

The goal here is to prevent us from getting merge conflicts each time we push to a PR because the JS gets rebuilt both locally, and on Federalist Build. We don't need to track our local builds because Federalist makes the compiled JS fresh each time.


.gitignore
```diff
@@ -7,5 +7,4 @@ node_modules
 .DS_Store
 playwright-report/
 test-results/
-playwright-report/
```
billyhunt, line 10: Was this supposed to be removed?

JackRyan1989: Yes because it's in there twice

billyhunt: Oh whoops!  Missed that.

billyhunt (approved):
Looks good!

c2nelson (approved):
thank you for the explanation and the fix!

PR #332, opened 2022-07-14
Parental rights overlay

c2nelson:
@vwilliamsn Take a look at the edits we discussed this morning.  Note that the link to the EC TA does not work.  We will have that page up before this one so it will work once it's live.


vwilliamsn:
@c2nelson - great job, this is looking really good. Can we check the font size on the accordion features.

vwilliamsn:
Looks good @c2nelson thanks for your work on this.  

2 quick questions:
1) Has the EC TA piece been fixed? 
2) Should I approve this now or wait until the picture has been added?

vwilliamsn (approved):
Great job, thanks @c2nelson

JackRyan1989:
@c2nelson Should the accordion that reads "Learn more about how the ADA protects others involved in the child welfare system who have a need for disability-related aids or services in order to communicate effectively—such as a relative or friend of a person seeking or receiving child welfare services." just be a link rather than an accordion? It's a really long title and the content within the accordion is very minimal.

PR #333, opened 2022-07-15
Moving Print Button JS over to wepback.

JackRyan1989:
The print-button.js script has been removed from the main.js file, and is now built by Webpack and pulled into the scripts.html template on it's own.

To Test:
1. Navigate to one of the pages in the 'Explore Featured Topics' section 
2. Click the "Print this page" button
3. Each page should print
4. Confirm there are no errors thrown to the console.

### DIFF
--- a/_assets/js/print-button.js
+++ b/_assets/js/print-button.js
@@ -1,14 +1,16 @@
-// Grab our print button:
-var printButton = document.getElementById('crt-page--printbutton');
-
-// Print function:
-function printPage() {
-    window.print();
-}
-
-// Assign an event listener - on click do this
-if (printButton) {
-    printButton.addEventListener('click', function(){
-        printPage();
-    });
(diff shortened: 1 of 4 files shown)


_assets/js/print-button.js
```diff
@@ -1,14 +1,18 @@
-// Grab our print button:
-var printButton = document.getElementById('crt-page--printbutton');
-
-// Print function:
-function printPage() {
-    window.print();
+const printButtonFunction = () => {
```
billyhunt, line 1: What do you think about changing this to "PrintButton" as its a component and function is implied?

billyhunt (approved):
Looks good to me.

_assets/js/print-button.js
JackRyan1989: Sure I'll change that back.

PR #351, opened 2022-08-08
Netlify Integration

billyhunt:
Add Netlify CMS to ADA.


billyhunt:
@JackRyan1989, I am getting a required field error when trying to save to the index page.

image

Should news be required?

JackRyan1989:
@billyhunt Good catch. Yeah, the News text should be required. It's the text for that section that displays when there are news items. Just wondering, but was there text in there already?

billyhunt:
It was not!

JackRyan1989:
@billyhunt Found the problem! The name field for the News widget was set to "News" instead of "news"

billyhunt:
Nice one!

JackRyan1989 (approved):
Looks good to me!

PR #352, opened 2022-08-08
Parking tag is now functional

JackRyan1989:
(Ticket)[https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/525]

## What does this change?

'parking' has been added as a tag to the data.yml and constants.js files to allow for tagging of parking related topics. 

## Checklist:

- [ ] Open the Preview link 
- [ ] Navigate to the resources page
- [ ] Confirm that only the Restriping Parking page appears when you select the 'Parking' tag from the drop down list.


c2nelson (approved):
thanks!

PR #354, opened 2022-08-10
Aria live region update for resources page

JackRyan1989:
I added aria-live=polite to explicitly make the live region on the Resources page polite. On mobile the screen reader has been announcing the number of search finds in the middle of other content on the page, I would like it to stop interrupting as we have it on desktop.


c2nelson (approved):
thx.

PR #372, opened 2022-08-19
Update optional fields add public directory for images

billyhunt:
Can you make sure the optional fields looks right to you and see that update and creates with images are working?


JackRyan1989 (approved):
The public directory works for images. 

The first screen shot here shows the image in the preview page, and the second screenshot shows the image appearing in the CMS.

Screen Shot 2022-08-22 at 7 36 43 AM
Screen Shot 2022-08-22 at 7 36 55 AM

PR #374, opened 2022-08-22
Medcare ta

c2nelson:


JackRyan1989:
@c2nelson the images that you have placed in tables, these are figures on the legacy site, right? We should make sure we adhere to a single design for figures. We should have @chinelo-ikejimba or @beckyslogeris confirm the design is ok because we'll need to copy that design on the Design Standards for those figures.

vwilliamsn:
Ok, please let me know when the image placement has been reviewed, so I can approve this pull request.

chinelo-ikejimba:
@c2nelson @vwilliamsn @JackRyan1989 @beckyslogeris 

Following notes on #375

Suggested design notes
Let's follow the figure and image working format in the 1991 Design Standards with the following guidelines:
- no border around figures/images
- caption placed in body text beneath the figure/image

vwilliamsn (changes requested):
This is looking great, thanks for your work on this @c2nelson.  

Two small changes - check the images under the following sections:
*Under Ceiling Mounted Lift
*Radiologic Equipment

c2nelson:
@vwilliamsn all set

vwilliamsn:
Looking good @c2nelson! 

Would you please check the image under 'Free Standing Overhead Lifts'.  I think this is the picture that should be there:
image

1. Free-standing overhead track lift systems function like ceiling-mounted lifts and do not require modifications to existing construction.
2. While not as portable as floor lifts with wheels, these lifts are movable and can be relocated as needs or services change.
3. An overhead lift does not require the additional maneuvering space needed by a portable floor lift.

vwilliamsn (approved):
Thanks for your work on this @c2nelson!

PR #381, opened 2022-09-02
Back to Top links for Design Standards pages - [WIP]

JackRyan1989:
### What does this change? 

This feature adds back to top button that appears in the bottom right hand side of the screen for mobile and tablet users.

1. Accessibility - flag for Terri to review after 1-3 is done. - APPROVED

### Testing
1. Confirm that the button only appears on the 1991 Design Standards page and no where else.
2. Confirm that the button only appears on the 1991 DS page on Tablet and Mobile viewports (<1024 px wide)
3. The button should not appear if the table of contents are in view
4. The button should not appear if you are initially scrolling down
5. The button should appear if you are below the TOC and are scrolling up, then should stay visible until you reach the top of the screen again.


c2nelson (commented):
Plugin looks good. This is going to be so useful!

beckyslogeris (commented):
@JackRyan1989 - Here's my 2 cents!

Overall it looks good (styling is correct & it's usable), but if we had more time I wonder if we could come up with a more elegant design solution.

My nit picky concerns:
- It feels redundant in places where the sections are shorter (or just a header)
- It doesn't help when you're lost in a sea of text for longer sections
- It might be annoying for screen readers (total assumption)

A few thoughts:
- Could we do a sticky "back to the top" button in the bottom right corner of the page?
[Screen Shot 2022-09-02 at 3 45 16 PM](https://user-images.githubusercontent.com/101133071/188226616-b9017c6a-64ef-47a8-9d62-6638f7a648a3.png)
- Wondering if the left hand navigation is enough?

All of that to say - I totally understand if there isn't time or if this isn't a priority. (Or if a decision was made in previous conversations that I missed!) Let me know if you want to chat at all next week.

JackRyan1989:
@beckyslogeris I'm starting to agree after building this out. My original hesitancy around a floating 'Back to Top' button was that it might not necessarily be the most accessible solution. Even though the button follows the user down the screen, it stays in the same place in the markup, so for screen readers its only in one spot, and it might be very far away in the markup from where they are.

That said, after adding a bunch of 'Back to Top' links, I think it might get really annoying for screen reader users. Especially if someone uses features that list all the links and headings. They'll see that link over and over again. 

I can make both, and we can directly compare them and have Terri weigh in about the accessibility of each solution.

beckyslogeris:
@JackRyan1989 - sounds good! I was also doing some research on screen readers and it seems like most of them have a built in "back to top" function - but Terri would definitely be the best person to ask!

JackRyan1989:
That totally makes sense! Thanks for looking into that @beckyslogeris

JackRyan1989:
@beckyslogeris Should the back to top button in your mock design always be visible? Even when the user is at the top of the page? Or should it only be visible when the user scrolls down?

beckyslogeris:
@JackRyan1989 - I think it should pop up when the user scrolls down. 

One idea from from NN/g is to even have it pop up only when someone indicates that they want to scroll up - "Consider delaying the appearance of a Back to Top option until after users scroll a few pages AND indicate that they want to scroll up (e.g., move the scroll bar up, flick up on the page).  Thus, the button appears only when it is most likely needed, without covering up important parts of the screen the rest of the time." [Back-to-Top Button Design Guidelines
](https://www.nngroup.com/articles/back-to-top/)

Thoughts?

JackRyan1989:
@beckyslogeris I like the guidelines. They intuitively make sense to me. I have done my best to implement the guidelines here. Check the functionality on tablet width, as there is some page content that overflows on mobile, which is another thing we should think about tackling later.

I left in the back to top links under each heading for comparison and for accessibility review, but I suspect they will be removed in favor of the button.

JackRyan1989:
@beckyslogeris can you check this URL and let me know if the button matches your UX suggestions? https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/jr/back-to-top-link/law-and-regs/design-standards/1991-design-standards/

Also let me know if you'd like to meet about this.

beckyslogeris:
@JackRyan1989 - I'm not seeing it on my end...

JackRyan1989:
@beckyslogeris you have to shrink the screen to a tablet width. One way to do this is to right-click on the screen and select 'Inspect', then click the small phone icon in the dev tools upper edge. 

https://developer.chrome.com/docs/devtools/device-mode/

beckyslogeris:
Gotcha, thanks! This is what it looks like...

[Screen Shot 2022-09-07 at 4 48 21 PM](https://user-images.githubusercontent.com/101133071/188975391-d785ecaa-457a-4440-9c8f-4ec6c67d38a8.png)

beckyslogeris:
@JackRyan1989 - got it! Looks great!

JackRyan1989:
Review by Terri complete. Button is approved. Next steps: remove Ruby code for back to top links. Get JavaScript reviewed by @billyhunt

JackRyan1989:
Note to self: Remove 1991 Design Standards page before merging.

_assets/js/utils/backToTop.js
```diff
@@ -0,0 +1,73 @@
...
+  button.classList.remove(noDisplay);
+};
+const hide = () => {
+  button.classList.add(noDisplay);
+  button.classList.remove(blockDisplay);
+};
+
+// Function to check scroll position, set flags for scrolling up, and for calling toggleShow
+const checkYPos = () => {
+  console.log('checking pos...')
```
billyhunt, line 36: Should be removed?

_assets/js/utils/backToTop.js
```diff
@@ -0,0 +1,73 @@
...
+const button = document.getElementById(buttonId);
+// Class Names:
+const blockDisplay = 'mobile:display-block';
+const noDisplay = 'mobile:display-none';
+// Target element for intersection observer:
+const topElement = 'crt-page--sidenav';
+const observerTarget = document.getElementById(topElement);
+// Observer Options:
+const options = {
+  root: null,
```
billyhunt, line 19: Whats this for?

_assets/js/utils/debounce.js
```diff
@@ -0,0 +1,13 @@
+// Adapted from Josh W. Comeau: https://www.joshwcomeau.com/snippets/javascript/debounce/
+
+const debounce = (callback, wait) => {
+    let timeoutId = null;
```
billyhunt, line 4: do you need this variable?

billyhunt (commented):
Just a couple of tiny things.  Otherwise looks really good!

_assets/js/utils/backToTop.js
JackRyan1989: Yes! Good catch.

_assets/js/utils/backToTop.js
JackRyan1989: The options are for the intersection observer. They define what to observe, root:null means check for intersection with the entire viewport, root margin means don't add any margin around the target element, and threshold of 1 means make sure 100% of the target intersects with the viewport before firing the callback.

_assets/js/utils/debounce.js
JackRyan1989: Yes, because it used to establish a timeout before firing the callback. See line 7.

billyhunt (approved):
Looks good to me!

PR #385, opened 2022-09-06
New markdown widget for creating html figures

JackRyan1989:
### What does this change?

This adds a new mark down widget that allows users to add in figures and figure captions.
You can still add figures using standard HTML. But this syntax is a little shorter. 
To use it:

{% figure %}
This can be whatever figure content you want. Add images like so:
{% assets path-to-image alt="alt text" %}
{% endfigure %}

If you want to add a caption, do it like so:

{% figure %}
This can be whatever figure content you want. Add images like so:
{% assets path-to-image alt="alt text" %}
{% figcaption %}
Some text!
{% endfigcaption %}
{% endfigure %}

If you want to add a title to your figure, add it like this:

{% figure MyTitle %}
This can be whatever figure content you want. Add images like so:
{% assets path-to-image alt="alt text" %}
{% figcaption %}
Some text!
{% endfigcaption %}
{% endfigure %}

### DIFF
--- a/_assets/js/utils/constants.js
+++ b/_assets/js/utils/constants.js
@@ -13,7 +13,8 @@ const TAGS = [
   'effective-communication',
   'parking',
   'medical-care',
-  'mobility-devices'
+  'mobility-devices',
+  'child-care'
 ];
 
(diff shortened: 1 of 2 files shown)


c2nelson (approved):
thumbs up!  thanks, jack.

PR #387, opened 2022-09-09
Role="status" not working in some areas - JS Fix [WIP]

JackRyan1989:
Role status is now working on the Guidance and Resources page. It will not work consistently on the search page as that page does a full refresh on search. 

### Testing:
1. Head to the Guidance and Resources page
2. Turn on MacOS voiceover - https://support.apple.com/guide/voiceover/turn-voiceover-on-or-off-vo2682/mac
3. Toggle the checkboxes
4. The Screen reader should announce the number of items returned.


_assets/js/pagination.js
```diff
@@ -25,12 +26,12 @@ function reqLoaded() {
...
     renderSearchPage(resJSON, urlParams, NUMBER_OF_RESULTS);
   } else {
-    renderSearchResults(`${noResults()}`, false);
+    target.innerHTML = totalResults(null, 'result');
   }
 }
 // If timeout:
 function reqTimeout() {
-  renderSearchResults(`${noResults()}`, false);
+  target.innerHTML = totalResults(null, 'result');
```
billyhunt, line 34: Should this be null or 0?

billyhunt (commented):
Looks good.  Just checking that one thing to make sure we are not type casting null to what should be 0.

_assets/js/pagination.js
```diff
@@ -1,7 +1,6 @@
-import renderSearchResults from './utils/renderSearchResults';
 import { NUMBER_OF_RESULTS, SEARCH_ENDPOINT, ACCESS_KEY, AFFILIATE } from './utils/constants';
 import renderSearchPage from './utils/renderSearchPage';
-import noResults from './templates/search/noResultsTemplate';
+import totalResults from "./templates/search/totalResultsTemplate";
```
billyhunt, line 3: Does noResultsTemplate still get used somewhere?

JackRyan1989: No, it should all be handled by the totalResultsTemplate.

_assets/js/pagination.js
JackRyan1989: I have totalResults set to handle 0 differently than null. Null is passed in when there is an server side error. Here if the request timesout, we pass in null to return the error text. If 0 is passed in, then that means there are no results, so we return the no results found text.

_assets/js/pagination.js
billyhunt: Cool!

_assets/js/pagination.js
billyhunt: I usually try and keep each variable as one data type.  I wonder if you could pass in -1 instead?  If you want to use null though, its cool.

billyhunt (approved):

_assets/js/pagination.js
JackRyan1989: That makes sense. I could pass in -1, I just need to alter the if checks because one is just checking if the number is less than 1. So I'll add one that checks explicitly for 0, and another that explicitly checks for -1.

PR #390, opened 2022-09-16
compressed image

c2nelson:
replaced image


c2nelson:
@vwilliamsn the original image was 10mb. I compressed it to 4

vwilliamsn:
@c2nelson that looks better, thanks for spotting that!

vwilliamsn (approved):
You rock @c2nelson!

PR #396, opened 2022-09-26
Sub navigation for long pages

JackRyan1989:
Changes:

1. Updated the jekyll subnav plugin to index h3s as well as h2s
2. Updated sidenav template to display subnavigation accordions
3. Added styling so they don't look like accordions elsewhere.

Putting this up so we can talk about this solution to navigating on pages with tons of content.


vwilliamsn (approved):
Love it - this is what I wanted... yay!!! Thanks for adding the sub navigation it makes the document much more usable. Thanks @JackRyan1989!!!

PR #399, opened 2022-10-04
Updated link text on laws and regs page for web accessibility

JackRyan1989:
@vwilliamsn


vwilliamsn (approved):
Thanks for acting so quickly @JackRyan1989!

PR #403, opened 2022-10-13
Updating the Configuration file to hide Netlify CMS admin page from Sitemap

JackRyan1989:
And also to remove the old search.gov configuration values. 

This change has no corresponding frontend changes.

To test:

- Run locally and head to localhost:4000/sitemap.xml
- Confirm that /admin is not in the sitemap.xml file
- Check that /admin is not in the sitemap.xml file at the Federalist Preview URL either

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -140,7 +140,7 @@ defaults:
       sidenav: true
       featured: false
   - scope:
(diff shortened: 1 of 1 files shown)


c2nelson (approved):
great, jack!  thanks.

PR #405, opened 2022-10-19
Removing google tag manager id from subscribe section

JackRyan1989:
Accidentally included an id used for Google Tag Manager in the subscribe button. Removing it to clean up analytics.


billyhunt (approved):
Looks good to me!

PR #415, opened 2022-11-10
Cleaning up a few links

JackRyan1989:
Found a few places where the title 2 and 3 primers were referenced that needed links.


c2nelson (approved):
thanks, jack.  good catch.

PR #418, opened 2022-11-14
Retagging and adding related content [WIP]

JackRyan1989:
@c2nelson I made a decision here that I want your input on.

We applied the title ii and title iii tags to all content on the site, and this was because those topics apply to everything. I think we should consider being more selective with these tags, since the idea of the tags is to determine relationships and exclude content.

So, I made a rough guess at which items should be related to each other by title ii and iii. We should walk through these changes together.


JackRyan1989:
Issues resolved in one on one with @c2nelson

c2nelson (approved):
Woot!  Thanks @JackRyan1989

PR #425, opened 2022-11-16
Update launch text in notice page

JackRyan1989:
@c2nelson @vwilliamsn Including the revised launch text about Technical Assistance docs.


vwilliamsn (approved):
Thanks @JackRyan1989!

PR #431, opened 2022-11-18
Fixed all link issues

JackRyan1989:
Went through and found all bum links. Most were links to the Regs. These have been fixed.

Also fixed small bug with the subnav accordions.


vwilliamsn (approved):
Thanks @JackRyan1989!

PR #432, opened 2022-11-21
Added word break to links & stacked table on Title II and Title II pages 

JackRyan1989:
Changes:
1. Added new class: .word-break, that on mobile screen sizes breaks the associated link word at any spot. This helps shrink link widths.
2. Changed Table so that it stacks on mobile screens. See screen shots. This uses the Responsive Stacked Table from USWDS: https://designsystem.digital.gov/components/table/

Old. Notice how the table overflows:

Screen Shot 2022-11-21 at 5 12 11 PM

New. Notice how the table stacks.

Screenshot_20221121-171049


vwilliamsn (approved):
Thanks @JackRyan1989!

PR #433, opened 2022-11-22
Updated node version for setup-node action to 16

JackRyan1989:
What's changed?

Our current node version specified for our build is 12. We want to upgrade to the latest which is 16. I specified in the linter.yml and ally.yml files to use setup-node with version 16 of node.


billyhunt (approved):
Looks good to me!

PR #437, opened 2022-12-01
Cleaning up google analytics

JackRyan1989:
I removed conflicting Google Analytics script tags and removed the existing configuration variables. Otherwise analytics is left as is.


billyhunt (approved):
Looks good to me!

PR #445, opened 2022-12-09
Trying to update site build process for md and css linting [WIP]

JackRyan1989:
Matched the build step for linter.yml to a11y.yml

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -52,7 +52,7 @@ jobs:
       - run: npm install --production
 
       - name: Build site
-        run: bundle exec jekyll build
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -49,10 +49,10 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: "16"
-      - run: npm install --production
+      - run: npm install
```
billyhunt, line 52: Do we want to install everything in dev, or just the linter?  We might want to move that package over instead.

JackRyan1989: So this installation is actually for the build process that happens below. At least that's what I think. And we need all that stuff for the CSS to build.

PR #448, opened 2022-12-16
Adding click tracking to search results

JackRyan1989:
What's changed?

We're adding click tracking to our search results so we can update search.gov with popularity scores for our search results.

This adds an onclick event for each link on the search results page and sends the following info to Search.gov:
- url clicked
- search query
- affiliate
- position
- module code
- access key

We aren't doing anything with the response at the moment.

No front-end design changes. 

To test: 
1. Fire up locally
2. Let me know so I open the search.gov click tracking admin console
3. Enter a search query
4. Check the console to make sure there are no errors logged
5. Click a link
6. Let me (Jack Ryan) know 
7. I'll confirm that the number of clicks listed in the admin console has increased by the number of links you've opened

### DIFF
--- a/_assets/js/click-tracking.js
+++ b/_assets/js/click-tracking.js
@@ -10,7 +10,7 @@ export default function clickTracking() {
   allLinks.forEach((link, index) => {
     const url = link.href;
     const query = encodeURI(getSearchParam('query'));
-    const position = Number(getSearchParam('offset')) + (index + 1);
+    const position = encodeURI(Number(getSearchParam('offset')) + (index + 1));
     // What we need to send to the click tracking endpoint:
     // url clicked
     // search query
--- a/_assets/js/click-tracking.js
+++ b/_assets/js/click-tracking.js
@@ -1,5 +1,6 @@
 // Import our constants from constandts.js
 import { CLICK_TRACKING_ENDPOINT, ACCESS_KEY, AFFILIATE, MODULE_CODE } from './utils/constants';
+import { getSearchParam } from './utils/searchParamUtils';
 
 export default function clickTracking() {
   // Get all the anchor tags that are children of b tags on the page:
@@ -8,16 +9,8 @@ export default function clickTracking() {
   // Loop through all the links and add a click event listener to each one:
   allLinks.forEach((link, index) => {
     const url = link.href;
-    const query = location.search.split('&').find((param) => param.startsWith('query='));
-    const position = location.search.split('&').includes('offset=') ?
-    Number(
-        location.search
-          .split('&')
-          .find((param) => param.startsWith('offset='))
-          .replace('offset=', '')
-      ) +
-      index +
-      1 : index + 1;
+    const query = encodeURI(getSearchParam('query'));
+    const position = Number(getSearchParam('offset')) + (index + 1);
     // What we need to send to the click tracking endpoint:
     // url clicked
     // search query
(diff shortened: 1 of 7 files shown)


_assets/js/click-tracking.js
```diff
@@ -0,0 +1,43 @@
...
+
+  // Loop through all the links and add a click event listener to each one:
+  allLinks.forEach((link, index) => {
+    const url = link.href;
+    const query = location.search.split('&').find((param) => param.startsWith('query='));
+    const position = location.search.split('&').includes('offset=') ?
+    Number(
+        location.search
+          .split('&')
+          .find((param) => param.startsWith('offset='))
```
okkays, line 16: For `param.startsWith('offset=')`, what happens if offset is the first param? 

e.g., `/some/route?offset=12&query=10`


I think a generalized `function getParams()` might be useful here, since I see we're splitting on '&' in a few places to achieve this.

Modern browsers have [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get#examples), but we could maybe implement something that takes `location.search` and returns a `Map([[paramOne, valueOne], [paramTwo, valueTwo]])` etc?

_assets/js/click-tracking.js
```diff
@@ -0,0 +1,43 @@
...
+          .replace('offset=', '')
+      ) +
+      index +
+      1 : index + 1;
```
okkays, line 20: Readability: This ternary statement is a bit hard to grok, wdyt about a `function getOffset()` that returns the offset or 0 if not found?

```suggestion
    const position = getOffset() + index + 1;
```

_assets/js/click-tracking.js
```diff
@@ -0,0 +1,43 @@
...
+    // What we need to send to the click tracking endpoint:
+    // url clicked
+    // search query
+    // affiliate
+    // position
+    // module code
+    // access key
```
okkays, line 27: Probably not, but do we need to wrap any of these in [encodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI)?

okkays (changes requested):

_assets/js/click-tracking.js
JackRyan1989: So I actually already have a function called getSearchParams already written, 🤦🏻 

The way it currently works is it takes the param you want to get, like query or offset, and returns the value in a string. Since it's a string, it does need to be URI escaped. Are you leaning towards using a map for any particular reasons?

_assets/js/click-tracking.js
JackRyan1989: Yes, we need to wrap the getSearchParam output in encodeURI because it return spaces between multi-word queries. It probably won't happen for the offset values, but I'll just do it anyway since it's returning a string.

_assets/js/click-tracking.js
okkays: Nah - I suggested a Map because it conforms more closely to URLSearchParams, but just grabbing each param from the string works, too.

_assets/js/click-tracking.js
JackRyan1989: As for affiliate, module code and access key, those never change so we shouldn't need to escape anything there.

JackRyan1989: And I'm gonna trust the urls in the link hrefs because they're provided by search.gov.

okkays (commented):
Code LGTM.

For testing it, did you want to find 10 minutes to meet and walk through it? Might be easier than going back and forth async.

JackRyan1989:
> Code LGTM.
> 
> For testing it, did you want to find 10 minutes to meet and walk through it? Might be easier than going back and forth async.

Yeah let's walk through together

okkays (approved):

PR #449, opened 2022-12-19
Bump ruby version and package-lock for M1 chips

okkays:
- 🌎 Newer macbooks have M1 chips
- ⛔ The M1 archictecture breaks a lot of older tech, like ruby 2.7.0
- ✅ This commit bumps the version up, fixing compilation on M1. There's also a change to package-lock to get the devserver's watch working on OS X.

Local smoke tests seems to work, but @JackRyan1989, can you let me know if there's additional testing I should do for this, since a version bump is pretty wide-reaching?


JackRyan1989 (approved):
Looks good to me! 

We can update Docker in a separate PR.

PR #450, opened 2022-12-19
Explore generating redirects for PDFs

okkays:
Note: this PR is based on #457, which adds regression tests.

Since we have access to archive ada, we can generate a list of existing content to redirect to from ada.

The list can be expanded to include more than what is in FILE_PATTERNS

This PR has ~three~ four! parts:

1. A script to generate the redirects.json from the archive repo
2. The generated redirects.json
3. A file to generate redirect HTML from a list of JSON entries
4. A javascript fallback for pdf files, etc, triggered on the 404 page

~This works locally, but it's worth testing out on prod to ensure cloud.gov pages picks up on the `pdf.html` file extension correctly.~ I can confirm that the html/htm redirects work. The other redirects don't seem to work in the federalist preview because of the relative url issue, but I have no reason to believe they won't work on production, and won't cause any harm if they don't.

Some documentation of this working, especially since the 404 page content has changed:

Assets (pdfs):

redirect-pdf

HTML and HTM files (we'll get a warning on generation if there's conflicts, here, as we did for index.html and 404.html. See manual_redirects.json for how that's overridden).

redirect-htm

And, finally, on slow 3G speeds, we display that we're redirecting, instead of displaying a blank page for 30 seconds before redirecting:

redirect-3g

Also worth showing, a true 404 does 404:

image


_plugins/redirect_json.rb
```diff
@@ -0,0 +1,43 @@
...
+module RedirectsJson
+  class RedirectsJsonPage < Jekyll::Page
+    # Use Jekyll's native absolute_url filter
+    include Jekyll::Filters::URLFilters
+
+    def initialize(site, from, to)
+      from += '.html' unless from.end_with?('.html')
+
+      @site = site
+      @name = from
```
JackRyan1989, line 17: Why are you creating an instance variable from the input from? It also looks like @name is never referenced after being processed?

okkays: There are things that Jekyll + plugins make use of internally to get information about pages. [Here's the source](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/page.rb#L33-L36), and [here's how jekyll-redirect does it](https://github.com/jekyll/jekyll-redirect-from/blob/master/lib/jekyll-redirect-from/redirect_page.rb#L5)

Calling the builtin `initialize` for a page does _too much stuff_ (reading yaml, etc) so I set the necessary metadata here to avoid adding noops for all of those other methods.

JackRyan1989 (changes requested):
This looks great! I have one question about the redirect_json.rb plugin and one requested change: please add make_redirects.rb to the exclude list in the _config.yml so it isn't pushed into the site directory.

_plugins/redirect_json.rb
JackRyan1989: Ok yeah. Sorry I'm just confused because I'm not seeing the @name variable referenced here after being processed, like in the Jekyll docs? If the intent is to process the from url, process() mutates the data in place, so the processed value is not being used here.

JackRyan1989: All clear!

JackRyan1989 (approved):
Tested locally and on in preview link. Looks good to me!

okkays:
I think this is fixed now. I can't get test the pdf stuff on pages because the js assets 404.

Front end change to take a look at: I've changed the 404 page to say that we're trying to redirect. If we fail to find the file, or something else goes wrong, we switch to the regular 404 message.

billyhunt (approved):
This looks good to me.  I wonder if we should write a simple manual test plan to include for after we deploy when we get back in the new year.

JackRyan1989 (approved):
Lookin' good! Thanks Kevin!

okkays:
A post-deploy smoke test which should cover regressions and new cases:

- [ ] The app still works
  - Go to: https://www.ada.gov/index.html
  - Expect: ada.gov
- [ ] Markdown redirects still work
  - Go to: https://www.ada.gov/1991standards/1991standards-archive.html
  - Expect: https://www.ada.gov/law-and-regs/design-standards/1991-design-standards/
- [ ] HTM redirects work
  - Go to: https://www.ada.gov/yellocab.htm
  - Expect: https://archive.ada.gov/yellocab.htm
- [ ] Asset redirects work
  - Go to: http://www.ada.gov/pcatoolkit/chap1toolkit.pdf
  - Expect: PDF to load after a brief wait on the 404 page.

PR #451, opened 2022-12-19
Only include GA and DAP scripts on production

JackRyan1989:
Pretty straightforward. 

If the jekyll environment is development (the default), don't include the Google Analytics and Digital Analytics Platform scripts.

The reason we don't have to explicitly state the environment for production is that it is set here by Federalist: [https://github.com/cloud-gov/pages-build-container/blob/6f9c2a2ee010c33c290f706325848c5db7fbb503/src/steps/build.py#L405](https://github.com/cloud-gov/pages-build-container/blob/6f9c2a2ee010c33c290f706325848c5db7fbb503/src/steps/build.py#L405)


okkays (approved):
nice :+1:

PR #461, opened 2023-01-05
Netlify CMS preview styling

JackRyan1989:
This request does the following:

1. Write a custom script to pull in the latest hashed version of our CSS file
2. Registers that CSS file with our CMS so the styling is referenced in our preview templates.


admin/index.html
```diff
@@ -1,25 +1,37 @@
...
+        margin-left: 75px;
+        width: 150px;
+      }
+      p {
+        color: red;
+      }
+      button {
+        margin-top: 0px !important;
+      }
+    </style>
```
okkays, line 35: Do you foresee this expanding much? is it worth having a separate `admin.scss`?

okkays (approved):
One comment, otherwise LGTM

admin/index.html
JackRyan1989: No, this is just the styling for the CMS itself, which I don't see us editing.

c2nelson (approved):

PR #462, opened 2023-01-05
Netlify CMS preview styling JS

JackRyan1989:
This builds off of #461 

In this PR we are using a lot of custom JS to register custom templates with Netlify.

Most of it is fairly straightforward, though I do have to do some recursion to try to turn child elements into React JS elements.


_assets/js/netlify/topics/generateChildElements.js
```diff
@@ -0,0 +1,30 @@
+const notAllowed = ['em', 'strong', 'u', 'i', 'b', 'ul', 'blockquote', 'table', 'thead', 'td', 'tr','tfooter', 'tbody', 'br', 'input'];
```
okkays, line 1: why are these not allowed? is it worth adding a comment explaining?

_assets/js/netlify/topics/generateChildElements.js
```diff
@@ -0,0 +1,30 @@
...
+        )
+      }
+  } else if (!notAllowed.includes(item.nodeName.toLowerCase())) {
+    return h(
+      item.nodeName.toLowerCase(),
+      { "id": item.id.toString(), "className": item.className },
+      item.innerHTML
+    )
+  }
+    }
```
okkays, line 27: nit: fix indentation

_assets/js/netlify/topics/generateChildElements.js
```diff
@@ -0,0 +1,30 @@
+const notAllowed = ['em', 'strong', 'u', 'i', 'b', 'ul', 'blockquote', 'table', 'thead', 'td', 'tr','tfooter', 'tbody', 'br', 'input'];
+
+const generateChildElements = (html) =>
+  html.map((item) => {
```
okkays, line 4: this is a `map`, but not all code paths return a value.

this will lead to undefined values in the array - consider adding a filter before the map:

```
html.filter(item => {
  if (item.children && item.children.length > 0) return true;
  if (notAllowed.includes(item.nodeName.toLowerCase()) return false;
  return true;
})
```

_assets/js/netlify/topics/generateChildElements.js
```diff
@@ -0,0 +1,30 @@
...
+          generateChildElements([ ...Array.from(item.children)])
+        )
+      }
+  } else if (!notAllowed.includes(item.nodeName.toLowerCase())) {
+    return h(
+      item.nodeName.toLowerCase(),
+      { "id": item.id.toString(), "className": item.className },
+      item.innerHTML
+    )
+  }
```
okkays, line 26: This code is a bit confusing because there's a few things going on at once:
- Dealing with name, id, and className
- Dealing with item content (if allowed)
- Dealing with the item's children

wdyt about separating those into three steps, and then calling `h` on the result?

```suggestion
    const isItemAllowed = !notAllowed.includes(item.nodeName.toLowerCase());
    const itemHasChildren = item.children?.length > 0;

    const content = [];
    if (isItemAllowed) content.push(item.innerHTML);
    if (itemHasChildren) {
      content.push(generateChildElements([ ...Array.from(item.children)]));
    }
    // This should never happen with a filter, but to be safe!
    if (!content.length) return;
    
    return h(
          item.nodeName.toLowerCase(),
          { "id": item.id.toString(), "className": item.className },
          ...content,
    );
```

_assets/js/netlify/topics/generateChildElements.js
```diff
@@ -0,0 +1,30 @@
+const notAllowed = ['em', 'strong', 'u', 'i', 'b', 'ul', 'blockquote', 'table', 'thead', 'td', 'tr','tfooter', 'tbody', 'br', 'input'];
+
+const generateChildElements = (html) =>
+  html.map((item) => {
+    if (item.children && item.children.length > 0) {
```
okkays, line 5: are we allowed to use optional chaining?

```suggestion
    if (item.children?.length > 0) {
```

_assets/js/netlify/topics/makeBodyContent.js
```diff
@@ -0,0 +1,14 @@
+const md = window.markdownit();
+md.options['html'] = true;
+md.options['linkify'] = true;
+const parser = new DOMParser();
+// This function takes our body content as input, turns the markdown into HTML,
+// then loops over the html elements
+const makeBodyContent = (bodyContent) => {
+  let html = parser.parseFromString(md.render(bodyContent), 'text/html');
```
okkays, line 8: nit: const vs let, here and below

_assets/js/netlify/topics/makeBodyContent.js
```diff
@@ -0,0 +1,14 @@
...
+  let elements = html.children[0].children[1].childNodes;
```
okkays, line 9: `html.children[0].children[1].childNodes` is a bit cryptic, and chains like this make me worried that elements may not exist (e.g., `html.children[0]` is often `head`)

can you add an implementation comment explaining what this should be/why it's guaranteed to be that?

_assets/js/netlify/topics/makeBodyContent.js
```diff
@@ -0,0 +1,14 @@
+const md = window.markdownit();
+md.options['html'] = true;
+md.options['linkify'] = true;
+const parser = new DOMParser();
+// This function takes our body content as input, turns the markdown into HTML,
+// then loops over the html elements
+const makeBodyContent = (bodyContent) => {
```
okkays, line 7: nit: use function syntax?

```suggestion
function makeBodyContent(bodyContent) {
```

_assets/js/netlify/topics/makeBodyContent.js
```diff
@@ -0,0 +1,14 @@
+const md = window.markdownit();
+md.options['html'] = true;
+md.options['linkify'] = true;
+const parser = new DOMParser();
+// This function takes our body content as input, turns the markdown into HTML,
+// then loops over the html elements
```
okkays, line 6: I think this might be a case of [this anti-pattern](https://blog.codinghorror.com/coding-without-comments/)

_assets/js/netlify/topics/preview.js
```diff
@@ -0,0 +1,34 @@
...
+      const relatedContent = entry.getIn(['data', 'related-content']);
+      const pageData = {
+        title: title,
+        leadText: leadText,
+        body: bodyContentsArr,
+        publishDate: publishDate,
+        updatedDate: updatedDate,
+        print: print,
+        pdf: sideNavPDF,
+        relatedContent: relatedContent,
```
okkays, line 24:
```suggestion
        title,
        leadText,
        body: bodyContentsArr,
        publishDate,
        updatedDate,
        print,
        pdf: sideNavPDF,
        relatedContent,
```

any reason not to use the compact syntax?

_assets/js/netlify/topics/preview.js
```diff
@@ -0,0 +1,34 @@
...
+      }
+      return makeTemplate(pageData);
+    },
+  });
+
+  CMS.registerPreviewTemplate('topics', preview);
+  CMS.registerPreviewTemplate('resources', preview);
+  CMS.registerPreviewTemplate('resources', preview);
+  CMS.registerPreviewTemplate('law-and-regs', preview);
+  CMS.registerPreviewTemplate('pages', preview);
```
okkays, line 34:
```suggestion
  const pages = [
    'topics',
    'resources',
    'resources',
    'law-and-regs',
    'pages'
  ];
  pages.forEach(page => CMS.registerPreviewTemplate(page, preview));
```

happy either way with this:
  - splitting it to an array will help you more easily scan to see the duplicate `'resources'`.
  - Keep it as is will help if there's pages where you need to specify something other than `page, preview`

admin/index.html
```diff
@@ -1,25 +1,80 @@
...
\ No newline at end of file
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>ADA Content Manager</title>
+  </head>
+  <body>
+    <!-- Include the script that builds the page and powers Netlify CMS -->
+    <script src="https://unpkg.com/netlify-cms@^2.0.0/dist/netlify-cms.js"></script>
+    <script>
```
okkays, line 11: this script block is kind of complicated - can it go in its own file?

admin/index.html
```diff
@@ -1,25 +1,80 @@
...
+      // We need to grab out most current version of our CSS file and our custom JS. How can we do that?
```
okkays, line 12: What if you did:
```
---
layout: null
---
<!doctype html>
```
to tell jekyll to process this file, then you can use asset includes:

```
{% include scripts.html %}
{% asset styles.css %} {% asset vendor/uswds/uswds-init.min.js !type %}
```

okkays (changes requested):

admin/index.html
JackRyan1989: You mean the netlify cms package?

_assets/js/netlify/topics/makeBodyContent.js
JackRyan1989: Thanks! I default to let which is prolly backwards.

_assets/js/netlify/topics/generateChildElements.js
JackRyan1989: Totally allowed! I'm not used to using this operator.

_assets/js/netlify/topics/makeBodyContent.js
JackRyan1989: Does this make sense?

`The DOMParser returns a full HTML document, so grabbing the first child element actually grabs the <html> element because it is the first element under <!doctype> the second child element of the <html> is the <body> element. We aren't starting at the <html> element but rather the <!doctype> tag.`

_assets/js/netlify/topics/makeBodyContent.js
JackRyan1989: Ah. Yeah, I self-identify as a junior dev, so this tracks.

okkays: anything that requires anything more is probably over-engineering 🤷

admin/index.html
okkays: this comment was in reference to the css fetching code, which is gone now, so nevermind!

_assets/js/netlify/topics/makeTemplate.js
```diff
@@ -0,0 +1,46 @@
+import generateChildElements from './generateChildElements';
```
okkays, line 1: not a blocker, but want to note somewhere that we should look into whether JSX is usable

okkays (approved):
I know this is draft but LGTM - feel free to re-request later if you end up adding more to this

_assets/js/netlify/formatDates.js
```diff
@@ -0,0 +1,6 @@
+export default function formatDates(date) {
```
okkays, line 1: good foresight making this a module

in my experience apps usually end up with quick a few different date formats - it might be good to name this function `formatPublishDate` or something to leave room for `formatSomeOtherDate` in the future

_assets/js/netlify/topics/makeBodyContent.js
okkays: lgtm, thanks!

okkays (changes requested):
Making sure we're grabbing attributes for accessibility is my main thing. Otherwise just readability comments/suggestions.

_assets/js/netlify/formatDates.js
JackRyan1989: Good catch!

_assets/js/netlify/formatDates.js
```diff
@@ -0,0 +1,6 @@
...
+    const newDate = new Date(date.split(" ")[0].split("-"));
+    const options = { month: "long", day: "numeric", year: "numeric" };
+    const finalDate = new Intl.DateTimeFormat("en-US", options).format(newDate);
+    return `${finalDate}`;
```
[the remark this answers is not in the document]
JackRyan1989: Good catch! No it's not necessary. Just checked and the `format()` method returns a string.

_assets/js/netlify/makeTemplate.js
```diff
@@ -0,0 +1,46 @@
+import generateChildElements from './generateChildElements';
+import formatDates from './formatDates';
+
+const makeTemplate = (data) => {
+  const { title, leadText, body, publishDate, updatedDate, print, pdf, relatedContent } = data;
+  return h(
+    'div',
```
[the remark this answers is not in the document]
JackRyan1989: Yep Just got it working and pushed those changes.

_assets/js/netlify/generateChildElements.js
```diff
@@ -0,0 +1,20 @@
...
+    if (item.nodeType !== Node.ELEMENT_NODE) return;
+    const itemHasChildren = item.childNodes?.length > 0;
+    const content = [];
+    if (itemHasChildren) {
+      content.push(generateChildElements([...Array.from(item.childNodes)]));
+    }
+
+    return h(
+      item.nodeName.toLowerCase(),
+      { id: item.id.toString(), className: item.className },
```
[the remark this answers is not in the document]
JackRyan1989: I have added code that should handle all attributes.

_assets/js/netlify/generateChildElements.js
```diff
@@ -0,0 +1,25 @@
...
+    const content = [];
+    if (itemHasChildren) {
+      content.push(generateChildElements([...Array.from(item.childNodes)]));
+    }
+    const attrsObj = {};
+    if (item.attributes.length > 0) {
+      for (const key of Array.from(item.attributes)) {
+        attrsObj[key.name] = key.value;
+      }
+    }
```
okkays, line 17: Another way of writing this:

```suggestion
    const attrs = Object.fromEntries([...item.attributes].map(
      attr => [attr.name, attr.value]
    ));
```

image

okkays (approved):
One style option, but LGTM!

_assets/js/netlify/generateChildElements.js
JackRyan1989: This is a good call. Much simpler and removes a variable. Thanks for catching!

PR #465, opened 2023-01-09
Leave room for action buttons on the sidenav

okkays:
- 🌎 Pages have "download" and "print" buttons for users who don't want to view the pages online.
- ⛔ On pages where the TOC takes up the full height, users have to scroll to the bottom of large documents to see the buttons.
- ✅ This PR fixes that by having the sidenav give room when in desktop mode. Mobile mode is unaffected.
- 🔮 Future PRs will add links to download their PDF equivalents

Before:

image

After:

image

### DIFF
--- a/_assets/sass/custom/_sidenav.scss
+++ b/_assets/sass/custom/_sidenav.scss
@@ -7,6 +7,12 @@
 }
 
 #crt-page--sidenav {
+  @include at-media(desktop) {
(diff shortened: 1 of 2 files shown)


kewitham (commented):
Looks good to me!

JackRyan1989 (approved):
Looks good to me!

PR #466, opened 2023-01-09
Limit page width to improve readability.

okkays:
- 🌎 As pointed out by Tina, some pages on ada.gov feel difficult to read.
- ⛔ Her theory is that this is because the text width is too large
- ✅ This commit explores shrinking that width to 64rem (the "small screen" max width used throughout the site)
- 🔮 This is a draft to evaluate whether this would make pages more readable - feedback is encouraged!

Before: scanning the line:

> Knowing when the 1991 or the 2010 ADA Standards apply to buildings and facilities is important in determining if your building or facility complies with the ADA. The [ADA Requirements: Effective Date and Compliance Date guide](https://archive.ada.gov/revised_effective_dates-2010.htm) helps to explain which version of the ADA Standards to use and when.

is difficult because the eye has to travel far on large screens.

image

After feels easier on the eyes:

image

### DIFF
--- a/_pages/cases.md
+++ b/_pages/cases.md
@@ -14,6 +14,7 @@ redirect_from:
   - /enforce.htm
   - /enforce_activities/
 sidenav: false
+compact: true
 lead: |-
   The Department of Justice enforces the ADA through lawsuits and settlement agreements to achieve greater access, inclusion, and equal opportunity for people with disabilities.
 ---
--- a/_assets/sass/custom/_page.scss
+++ b/_assets/sass/custom/_page.scss
@@ -9,6 +9,13 @@
   padding-right: 2rem;
   width: 100%;
 
+  @include at-media(desktop) {
+    &.compact {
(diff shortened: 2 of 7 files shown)


JackRyan1989:
@okkays Initial run through with screen reader seems good to me. No discernible difference between this change and live site.

_pages/cases.md
```diff
@@ -14,6 +14,7 @@ redirect_from:
   - /enforce.htm
   - /enforce_activities/
 sidenav: false
+compact: true
```
JackRyan1989, line 17: Let's make sure this is documented in our wiki: https://github.com/usdoj-crt/beta-ada/wiki/Front-matter

JackRyan1989 (approved):
Nice. This looks great. No concerns!

_pages/cases.md
okkays: nice, thanks!

PR #469, opened 2023-01-11
Allow manually triggering workflows.

okkays:
- 🌎 These files run when a branch is pushed to the repository.
- ⛔ In some cases, they won't run - like if the user doesn't have access to the repo
- ✅ This commit lets folks with access trigger CI/CD against branches. It will also let us rerun failed actions (in the even of timeout, etc) without requiring a duplicate branch push.

See: https://github.blog/2022-02-10-using-reusable-workflows-github-actions/

FYI - would require sharing actions with other projects in usdoc-crt to make available:

image

### DIFF
--- a/.github/workflows/a11y.yml
+++ b/.github/workflows/a11y.yml
@@ -1,5 +1,7 @@
 name: Run Accessibility Tests
-on: [push]
+on:
+  push:
+  workflow_call:
 jobs:
   build:
     name: Run Accessibility Tests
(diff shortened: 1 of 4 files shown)


okkays:
Closing - turns out the issue isn't a permissions issue, it's that Katie is pushing to her own repo, not ours. So this might not actually solve the problem :D

We'll simply need to push her commits to branches that exist on beta-ada until she gets access.

PR #479, opened 2023-01-25
Fix compact alignment on large screens

okkays:
This fixes an issue on very large screens, where the alignment for the new compact pages was off (see GIFs below).

- 🌎 Compact pages don't stretch to fill the screen
- ⛔ On large screens, this means they are limited to the left 64rem of the screen
- ✅ This commit auto-aligns them in the center, in the same was as the toolbar.

Before:

(The misalignment is only on the large resolution - other resolutions are unaffected)

before

After:

(All resolutions look good now!)

sizes

### DIFF
--- a/_assets/sass/custom/_page.scss
+++ b/_assets/sass/custom/_page.scss
@@ -22,8 +22,7 @@
       margin-right: auto;
       max-width: 64rem;
 
-      position: relative;
-      right: calc((95rem - 64rem) / 2);
+
     }
   }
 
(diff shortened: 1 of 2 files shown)


vwilliamsn (approved):
@okkays - thanks for catching and fixing this issue.

okkays:
> @okkays - thanks for catching and fixing this issue.

this was @nguyen-tina's catch 😄

Definitely worth keeping in mind to test for:
- mobile
- desktop (small) [< 64rem]
- desktop (large) [> 64rem]

Once CCI passes with this last tweak and I double check pages, I'll go ahead and merge this

PR #486, opened 2023-01-31
Fix page offset at narrow resolutions

okkays:
This is a quick followup to the previous PR (https://github.com/usdoj-crt/beta-ada/pull/479) which missed an issue that affected the compact view at narrow resolutions:

- 🌎 We try to align compact pages with the title bar
- ⛔ Right now this is causing the page to move off of the screen on specific (narrow, desktop) resolutions, as we're using position: relative with a definite right margin with certain widths.
- ✅ This commit changes the approach to force the page to the same width as the title bar at all times (95rem) but with padding to align it properly

Before:

before-margin

After:

after-margin

### DIFF
--- a/_assets/sass/custom/_page.scss
+++ b/_assets/sass/custom/_page.scss
@@ -16,14 +16,19 @@
     }
   }
 
-  @media(min-width: 64rem) {
+  @media(min-width: 95rem) {
     &.compact {
       margin-left: auto;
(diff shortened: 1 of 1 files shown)


nguyen-tina (approved):
Looking good!

kewitham (approved):
lgtm!

PR #488, opened 2023-01-31
Optimize Images

okkays:
This is a followup to introducing the new Design Standards images for https://github.com/usdoj-crt/beta-ada/pull/487

In that PR, I noticed our image assets were relatively unoptimized, so I ran a utility to make some quick improvements. Further improvements could be made wherever we don't mind sacrificing image resolution or marginal quality (for example, on the topics page).

## Why does this matter?

Pages with large images (such as the topics page) can take a long time to load, especially over 3G or satellite internet connections. These images also take valuable bandwidth - both for users with monthly bandwidth limits (mobile data and some desktop users), and ada.gov servers.

*Note: this GIF taken _after_ compression*

slow

## How to review this:

(I also did this for all images myself, but would appreciate the sanity check)

For each image attached:

- Confirm that the new image is smaller than the old one. This should be red and show a negative number: image
- Confirm that the modified image looks the same as the old one (there's no loss in color, quality, etc). You can click "Swipe" to get a larger view: image
- If the image shows "unable to render rich display", click "display the source diff" on the right to re-render it: image

## What does this change?

- 🌎 Images can often be compressed to take up less space without
  looking worse. Compressed images:
  - Download faster, serving users with slower connection speeds better
  - Take up less bandwith to deliver
- ⛔ A bunch of the ADA images showed significant size decrease with effectively lossless optimization.
- ✅ Running a compression utility (`jpegoptim` and `optipng`) on our ADA images resulted in
  significant gains in a few places

In total, the `_assets/images` was reduced from 38MB to 14MB:

Before: 

```shell
└─ $ du -h _assets/images
29M     _assets/images/landing
5.4M    _assets/images/project-images
2.6M    _assets/images/design-standards
38M     _assets/images
```

After:

```
└─ $ du -h _assets/images
5.1M    _assets/images/landing
4.5M    _assets/images/project-images
2.6M    _assets/images/design-standards
14M     _assets/images
```


c2nelson:
Yes!  Thank you, @okkays.  This was much needed!

kewitham (approved):
lgtm!

PR #491, opened 2023-02-09
pool-requirements-ta

c2nelson:
pool requirements TA and added anchors to DS


vwilliamsn:
Great job @c2nelson - add print button, 2 small changes, which are in bold. [Sidebar: The document had these errors in them, this has nothing to do with]

If you have purchased a non-fixed lift before **March15 th** that otherwise complies with the requirements in the 2010 Standards for pool lifts (such as seat size, etc.), you may use it, as long as you keep it in position for use at the pool and operational during all times that the pool is open to guests. Because of a misunderstanding by some pool owners regarding whether the use of portable pool lifts would comply with barrier removal obligations, the Department, as a matter of prosecutorial discretion, will not enforce the fixed elements of the 2010 Standards against those owners or operators of existing pools who purchased portable lifts prior to March 15, 2012 and who keep the portable lifts in **positon** for use at the pool and operational during all times that the pool is open to guests so long as those lifts otherwise comply with the requirements of the 2010 Standards. Generally, lifts purchased after March 15, 2012 must be fixed if it is readily achievable to do so.

vwilliamsn (changes requested):
Check hyperlink - otherwise everything looks good!

vwilliamsn (approved):
You rock @c2nelson!

PR #516, opened 2023-03-02
Fix Broken OUD Links

okkays:
The OUD topic has broken links to a PDF. This fixes them to point to OUD guidance, instead.


vwilliamsn (approved):
Thanks @okkays for the quick response!

kewitham (approved):

PR #569, opened 2023-04-04
[Netlify] Render Includes components without variables

okkays:
cc: @vwilliamsn @c2nelson 

# What is this?

This adds support for rending HTML includes.

# Limitations

It turns out a lot of the includes depend on _site variables_, such as what other posts exist and what those posts include. For example, to render news items, you need to know what news pages exist. That's something netlify doesn't know about, because netlify only cares about the current page.

Because those exist in the Jekyll context prior to loading our site, it's not possible (or at least, it's very difficult) to include a version of those variables for Netlify to parse. This means that includes which depend on other components (like the news feature) will fail to load.

Some of the listed includes also depend on _layouts_ being supported, but we don't support them yet. The includes themselves should work, but we can't actually preview them in netlify to check (e.g., back to top)

# Proof it works:

Here's a list of the includes from the ticket as AC's:

- [x] landing hero

hero

- [x] landing learn

learn

- [x] landing report

report

- [x] landing service-animals

Note: images are broken pending future work.

image

- [ ] news-item

This turned out to be hard, as it depends on other site content to render.

- [ ] card

This turned out to be hard, as it depends on "topics" existing on the site object.

- [x] download-pdf

image

- [x] expand-accordions-button
- [x] expand-accordions-wrapper

Note that for these, the 'expandable' text is due to the not-yet-implemented "details" widget:

expand

- [x] print-button

image

- [ ] back-to-top

This depends on us rendering `layouts` for the pages.

### DIFF
--- a/admin/config.yml
+++ b/admin/config.yml
@@ -43,7 +43,7 @@ collections:
   # Pages:
   - label: Pages
     name: pages
-    summary: "{{filename}} ({{title}})"
+    summary: "{{title}} ({{filename}})"
     files:
     # 404 error page:
       - label: "404"
--- a/admin/config.yml
+++ b/admin/config.yml
@@ -43,6 +43,7 @@ collections:
   # Pages:
   - label: Pages
     name: pages
+    summary: "{{filename}} ({{title}})"
     files:
     # 404 error page:
       - label: "404"
@@ -279,7 +280,7 @@ collections:
           widget: markdown
       # Resources page:
       - label: Resources
-        name: Resources
+        name: resources
         file: _pages/resources.md
         fields:
         - label: Permalink
@@ -298,6 +299,7 @@ collections:
     # Topic pages:
   - label: Topics
     name: topics
+    summary: "{{filename}} ({{title}})"
     folder: _pages/_topics
     create: true
     fields:
@@ -368,6 +370,7 @@ collections:
   # Resources pages:
   - label: Resources
     name: resources
+    summary: "{{filename}} ({{title}})"
     folder: _pages/_resources
     create: true
     fields:
@@ -452,6 +455,7 @@ collections:
     # Laws and regulations pages
   - label: Law and Regs
     name: law-and-regs
+    summary: "{{filename}} ({{title}})"
     folder: _pages/law-and-regs
     create: true
     fields:
@@ -488,6 +492,7 @@ collections:
   # Design standards pages
   - label: Design Standards
     name: design-standards
+    summary: "{{filename}} ({{title}})"
     folder: _pages/law-and-regs/design-standards
     create: true
     fields:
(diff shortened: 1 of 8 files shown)


kewitham:
@okkays a bit confused about the inclusion of the expand-accordions-button and expand-accordions-wrapper here because the details widget was working like this in the PR this is stacked on.

admin/config.yml
```diff
@@ -43,6 +43,7 @@ collections:
   # Pages:
   - label: Pages
     name: pages
+    summary: "{{filename}} ({{title}})"
```
kewitham, line 46: This is just a style thing but I think we should either keep the title as is and include the filename as a field in the entry page or have title first and filename in parentheses.

okkays:
> @okkays a bit confused about the inclusion of the expand-accordions-button and expand-accordions-wrapper here because the details widget was working like this in the PR this is stacked on. Ditto for the print button although did you get it to actually function? It's not working for me if so.

@kewitham I didn't think any `include`s were functioning before this. For example, this is how we load those accordions in the code: `{% include expand-accordions-wrapper.html expand-heading='Coverage' %}`. This shouldn't be related to your widgets work, right? I think the details widget looks the same but is a different component entirely, e.g.: `{% details Ramps expandable %}`

Re: function of things - nothing is going to function properly without its associated javascript. I don't know how the print function would even behave on this page :D

My goal was to render the includes, which does exclude the javascript. It might be interesting to try to load the javascript for the site, but we'll probably want to handle that in a followup PR - wdyt?

admin/config.yml
okkays: I had it this way because a lot of the titles are fairly long, but happy to switch it around (done)

okkays: (imo including the filename gives a unique, concise point of reference between developers and content authors)

kewitham:
> 
I thought this note in the PR description was referring to the details widget - **Note that for these, the 'expandable' text is due to the not-yet-implemented "details" widget** - is that referring to something else?
The print button comment is because you have it here as part of the PR description but it was something that was already rendering.

okkays:
@kewitham ah, I see

The PR description came from the list in the issue (717) - I was checking off whether they render with this include code.


## Details widget

So these details accordions are made up of a few pieces:
- expand-accordions-wrapper (an include, which this PR renders but has no javascript)
- expand-accordions-button (an include, which this PR renders but has no javascript)
- the details widget (which you're right, you've implemented, but I think might have a bug:)

I think there's a bug in the details widget where it shows "expandable" as text - this is [something that's captured here in the ruby plugin](https://github.com/usdoj-crt/beta-ada/blob/kk/717/netlify-includes/_plugins/details.rb#L7-L9).

Clicking the expand/collapse button doesn't work (and won't without javascript) but the button renders and can be "proofed"

image

Anyway, I think we're good to go on the accordions front.

## Print button

Print button renders (and is in the _includes folder - so I didn't think your work implemented it?) either way it's good to go 😅

kewitham:
@okkays makes sense and lgtm!

kewitham (approved):

PR #576, opened 2023-04-12
Fix netlify color contrast issues

kewitham:
This PR addresses the color contrast issues outlined in [Issue #773](https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/733).

**Before**
Screenshot 2023-04-12 at 3 29 20 PM

**After**
Screenshot 2023-04-12 at 3 32 23 PM

**Before**
Screenshot 2023-04-12 at 3 29 58 PM

**After**
Screenshot 2023-04-12 at 3 32 43 PM

**Before**
Screenshot 2023-04-12 at 3 30 25 PM

**After**
image

**Before**
Screenshot 2023-04-12 at 3 34 15 PM

**After**
Screenshot 2023-04-12 at 3 34 04 PM

### DIFF
--- a/admin/index.html
+++ b/admin/index.html
@@ -30,6 +30,18 @@ layout: null
       button {
         margin-top: 0px !important;
       }
+      .css-g42qkj-ToolbarSubSectionFirst .css-16pehfe-StyledDropdownButton-button-default-caret-caretDown-DropdownButton-noOverflow-PublishedToolbarButton-publishedButton  {
+        color: #117888;
+      }
+      .css-g42qkj-ToolbarSubSectionFirst .css-1ixjrmh-ToolbarButton-button-default-buttonMargin-noOverflow-DeleteButton-lightRed {
+        color: #D60032;
+      }
+      .css-1prcas7-ControlPaneContainer .css-pcief1-FieldLabel-fieldLabel {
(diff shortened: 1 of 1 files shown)


admin/index.html
```diff
@@ -30,6 +30,18 @@
       button {
         margin-top: 0px !important;
       }
+      .css-g42qkj-ToolbarSubSectionFirst .css-16pehfe-StyledDropdownButton-button-default-caret-caretDown-DropdownButton-noOverflow-PublishedToolbarButton-publishedButton  {
```
okkays, line 33: What is `g42qkj`? It looks like some kind of generated hash - is this going to change if we update netlify, for example?

okkays (changes requested):

admin/index.html
kewitham: @okkays I've been playing around with how to do this without these classes and there's not a great way that I could find. I'm thinking we go with this in the short term and I will create a PR against the Netlify repo to see if we can get a more permanent fix implemented.

okkays: sgtm!

okkays (approved):
lgtm given this will eventually need to come from upstream.

PR #583, opened 2023-04-19
Add preview link on Netlify

kewitham:
This PR surfaces the federalist preview link in Netlify so users can see their changes on the site without going to GitHub.

image


_assets/js/netlify/getPreviewLink.js
```diff
@@ -0,0 +1,16 @@
...
+        link = entry.getIn(['data', 'permalink']);
+    }
+    const status = entry.get('status');
+
+    if (status === '') {
+        return null;
+    }
+
+    return 'https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/cms'.concat(link);
+}
```
okkays, line 16: I see three cases to this - below might be a more concise way to express this?
- No status -> null
- Is a card -> card href
- Else -> permalink

Is it safe to assume permalink exists?

```suggestion
const URL = 'https://federalist-1716bf9c-dd79-4d96-8285-6e56dc391b84.app.cloud.gov/preview/usdoj-crt/beta-ada/cms';

export default function getPreviewLink(entry) {
  if (entry.get('status') === '') return null;

  const card = entry.getIn(['data', 'card']);
  if (card) return URL + card.get('href');
    
  const permalink = entry.getIn(['data', 'permalink'])
  if (permalink) return URL + permalink;
  
  console.error('Unable to build pages from entry', entry);
  return null;
```

_assets/js/netlify/makeTemplate.js
```diff
@@ -4,12 +4,17 @@ import formatPublishDates from './formatPublishDates';
...
   const pageTemplate = (
     <div
       id="crt-page--content"
       className="mobile-lg:grid-col tablet:grid-col-12 desktop:grid-col-8"
     >
+      {previewLink ? (
+        <div className="width-full padding-bottom-2 padding-top-2 border-bottom">
+          <a href={previewLink} target="_blank" className="margin-top-4 text-bold line-height-mono-6">Page preview link</a>
+        </div>
+      ) : null}
```
okkays, line 17: is it worth adding a "Move to review to get a preview link" instead of hiding this completely?

(Users might be confused by its absence if they expect a link to exist / don't understand how/when the link is created)

okkays (changes requested):
Nice. A couple of comments, but super clean way of adding this 👍

_assets/js/netlify/makeTemplate.js
kewitham: good call, added!

okkays (approved):
LGTM. Not sure why pages build isn't running :/

vwilliamsn (approved):
Thanks for all your hard work @kewitham!

PR #597, opened 2023-05-01
Fix Netlify error when page is unexpected.

okkays:
Technical issue - this was causing an exception message in the chrome console, but didn't seem to be causing errors otherwise.

- 🌎 This code checks for `currentPage[6].concat`, trying to find a collection name.
- ⛔ If `currentPage[6]` is undefined, that throws an exception
- ✅ This commit adds a check to make the expression undefined if `currentPage[6]` is undefined.


kewitham (approved):
thanks for fixing! lgtm!

PR #608, opened 2023-05-10
Don't stop rendering on template failures.

okkays:
- 🌎 The netlify system breaks entirely if a template is invalid.
- ⛔ This isn't great because as you're typing a template out, it's invalid.
- ✅ This commit catches those exceptions and displays them nicely.

example


kewitham (approved):
lgtm!

PR #609, opened 2023-05-10
Turn off pa11y color-contrast

okkays:
We've been getting too many false alarms. I'd previously turned off mobile, but desktop is acting up, too.

See: github.com/pa11y/pa11y/issues/633


kewitham (approved):
lgtm!

PR #628, opened 2023-06-01
Fix new image loading and detail component

kewitham:
This PR addresses a couple issues in Netlify:
1. When images were added in new PRs they were not rendering properly. This PR adds logic to get the paths for those images.
2. The detail includes component was not rendering properly where the text contained apostrophes. This PR adds logic to remove those so the components can be parsed properly and then add them back once they are parsed.


c2nelson (approved):
nice!  thank you.

PR #632, opened 2023-06-16
Homepage notice fix

kewitham:
This PR updates the notice component to only show the h3 tag if there is notice text

### DIFF
--- a/_includes/notice.html
+++ b/_includes/notice.html
@@ -5,7 +5,9 @@
   <div class="usa-alert">
     <div class="usa-alert__body">
       {% for post in site.categories.notices reversed limit:1 %}
-      <h3 class="usa-alert__heading pa11y-skip">{% if lang=="en" %}{{ post.notice_text }}{% elsif lang=="es" %}{{post.notice_text_es}}{% endif %}</h3>
+      {% if post.notice_text != nil %}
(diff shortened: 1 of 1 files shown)


_includes/notice.html
```diff
@@ -5,7 +5,9 @@
   <div class="usa-alert">
     <div class="usa-alert__body">
       {% for post in site.categories.notices reversed limit:1 %}
-      <h3 class="usa-alert__heading pa11y-skip">{% if lang=="en" %}{{ post.notice_text }}{% elsif lang=="es" %}{{post.notice_text_es}}{% endif %}</h3>
+      {% if post.notice_text != nil %}
```
okkays, line 8: Is the comparison to nil necessary (could this be `{% if post.notice_text %}`)?

Also, what if `post.nostice_text_es` is not nil?

Sidenote: There are a couple of places in this code where the pattern `{% if lang=="en" %}{{ post.notice_text }}{% elsif lang=="es" %}{{post.notice_text_es}}{% endif %}` is referenced. Could this be made [a variable?](https://shopify.github.io/liquid/tags/variable/)

okkays (changes requested):

okkays (approved):

PR #635, opened 2023-06-21
Add new accessible laws and regs PDFs

okkays:
- ⛔ The laws and regs were missing accessible PDF downloads.
- ✅ This commit adds the new PDFs in a few locations:
  - The law-and-regs landing page: landing page example
  - The title-ii and title-ii pages, as "download pdf guidance": download button example


vwilliamsn:
@okkays thanks for getting this done so quickly!  I'm not sure if others are having a problem but the federalist link won't let me download the PDF.  

image

okkays:
@vwilliamsn I think that's expected, unfortunately, due to the way pages preview works with URLs

The design standards URLs also don't work, for example

vwilliamsn (approved):
Thanks for the clarification @okkays! LGTM!

PR #636, opened 2023-06-21
Update Ruby and Gems (2.7.8)

okkays:
- 🌎 Ruby is now on 2.7.8, which includes security fixes.
- ⛔ We are on ruby 2.7.4
- ✅ This commit upgrades to 2.7.8

Developer note: 2.7.8 also compiles on an M1 chip, so compatibility mode
can be disabled.


kewitham (approved):
lgtm!

c2nelson (approved):

PR #637, opened 2023-06-21
Speed up development by disabling redirects

okkays:
Note: This only affects the development workflow.

- 🌎 We build the entire site when live reloading in development
- ⛔ Right now this takes ~50 seconds (for me).
- ✅ This commit removes redirects (generated and the redirect
  directory) from dev compilation, which speeds up compilation to ~16
  seconds)
- 🔮 Future commits will further debug our sass compilation time, which
  is 8 of the remaining 16 seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ This site requires Ruby and Node.js in order to run locally. For guidance on ins
     $ bundle install
     $ npm start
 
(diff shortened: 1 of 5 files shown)


kewitham (approved):
lgtm!

vwilliamsn (approved):
Thanks @okkays for your work on this!

c2nelson (approved):

PR #638, opened 2023-06-23
ballot box accessibility build along with pdf for download

c2nelson:
ballot box accessibility build along with pdf for download


vwilliamsn (approved):
This looks great, thanks @c2nelson!

PR #639, opened 2023-06-23
Add expand all button to sidenav

kewitham:
This PR adds an expand all button to the sidenav to enable viewing all the submenus at once.

image
Screenshot 2023-06-23 at 12 54 48 PM

### DIFF
--- a/_assets/js/expand-sidenav.js
+++ b/_assets/js/expand-sidenav.js
@@ -1,0 +1,32 @@
+const sidenav = () => {
+    const expandButton = document.getElementById('expand-all');
+
+    function expandAll(e) {
+        const submenuButtons = document.getElementsByClassName('usa-accordion__button');
+        if (e.target.innerText === 'Expand all') {
+            Array.from(submenuButtons).forEach(button => {
+                const isExpanded = button.getAttribute('aria-expanded') === "true";
+                if (!isExpanded) {
+                    button.click();
+                }
+            });
+            e.target.innerText = 'Collapse all';
+        } else {
+            Array.from(submenuButtons).forEach(button => {
+                const isExpanded = button.getAttribute('aria-expanded') === "true";
+                if (isExpanded) {
+                    button.click();
+                }
+            });
+            e.target.innerText = 'Expand all';
+        }
+    }
+
+    if (expandButton) {
+        expandButton.addEventListener('click', function(e){
+            expandAll(e);
+        });
+    }
+}
+
+export default sidenav;
--- a/_assets/js/expand-sidenav.js
+++ b/_assets/js/expand-sidenav.js
@@ -3,23 +3,13 @@ const sidenav = () => {
 
     function expandAll(e) {
         const submenuButtons = document.getElementsByClassName('usa-accordion__button');
-        if (e.target.innerText === 'Expand all') {
-            Array.from(submenuButtons).forEach(button => {
-                const isExpanded = button.getAttribute('aria-expanded') === "true";
-                if (!isExpanded) {
-                    button.click();
-                }
-            });
-            e.target.innerText = 'Collapse all';
-        } else {
-            Array.from(submenuButtons).forEach(button => {
-                const isExpanded = button.getAttribute('aria-expanded') === "true";
-                if (isExpanded) {
-                    button.click();
-                }
-            });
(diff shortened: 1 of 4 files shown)


_assets/js/expand-sidenav.js
```diff
@@ -0,0 +1,32 @@
...
+            e.target.innerText = 'Collapse all';
+        } else {
+            Array.from(submenuButtons).forEach(button => {
+                const isExpanded = button.getAttribute('aria-expanded') === "true";
+                if (isExpanded) {
+                    button.click();
+                }
+            });
+            e.target.innerText = 'Expand all';
+        }
```
okkays, line 22: Suggestion: It looks like you can [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this up a bit:

```suggestion
        const submenuButtons = document.getElementsByClassName('usa-accordion__button');
        expandingAll = e.target.innerText === 'Expand all';
        Array.from(submenuButtons).forEach(button => {
            const isExpanded = button.getAttribute('aria-expanded') === "true";
            if (!isExpanded && expandingAll) button.click();
            if (isExpanded && !expandingAll) button.click();
        });
        e.target.innerText = expandingAll ? 'Collapse all' : 'Expand all';
```

_includes/expand-button.html
```diff
@@ -0,0 +1 @@
+<button style="width: max-content" class="usa-button desktop:margin-top-3 margin-bottom-2 bg-primary-darker" id="expand-all">Expand all</button>
```
okkays, line 1: cc @nguyen-tina, but I would change this button color to match the headers, something like:

image

_assets/js/expand-sidenav.js
```diff
@@ -0,0 +1,32 @@
+const sidenav = () => {
+    const expandButton = document.getElementById('expand-all');
+
+    function expandAll(e) {
+        const submenuButtons = document.getElementsByClassName('usa-accordion__button');
+        if (e.target.innerText === 'Expand all') {
+            Array.from(submenuButtons).forEach(button => {
+                const isExpanded = button.getAttribute('aria-expanded') === "true";
+                if (!isExpanded) {
+                    button.click();
```
okkays, line 10: Because we're simulating a click of each button, does this cause lag on less performant devices?

_layouts/page.html
```diff
@@ -23,6 +23,7 @@ <h4 class="usa-accordion__heading desktop:display-none">
       {% include sidenav.html collection=page.subnav %}
       </div>
       {% endif %}
+      {% include expand-button.html %}
```
okkays, line 26: Does/should this show up on mobile?

okkays (changes requested):
nice work! a few style changes / couple of questions

_includes/expand-button.html
kewitham: The button matches the print button below it right now so unless we're going to change the color of that I think it might look strange to have it a slightly different blue.

_layouts/page.html
kewitham: It does and I think it should since a lot of users are on mobile devices and may want to use it cc: @nguyen-tina

_assets/js/expand-sidenav.js
kewitham: I've only been able to test on my laptop but it happens instantaneously so I think it should be fine?

_includes/expand-button.html
okkays: I think the `+` is `#000` right now, and the print button is `#172E51`?

black

_layouts/page.html
okkays: sorry, you're right - for some reason I was under the impression that the mobile menu was pre-expanded, but it's definitely not.

_includes/expand-button.html
kewitham: oh you're talking about the +s? I'm not sure why those are black but I didn't touch those as part of this change so I'm inclined to leave them as is

nguyen-tina: I'm with Katie on that – I don't think the +'s are part of this, and am also inclined to leave them as-is

okkays: sgtm!

okkays (approved):

PR #643, opened 2023-06-29
Make Download PDF text configurable

okkays:
# What does this change?

Re: https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/813#issuecomment-1612991724

- 🌎 Right now the button can only say "Download PDF Guidance"
- ⛔ We want to make buttons for non-guidance downloads
- ✅ This commit enables that, and adds flags to edit it in netlify

# Proof it works

## Laws and Regs:

image

## Guidance:

image


kewitham (approved):
lgtm!

c2nelson (approved):
thank you!  looks good.

PR #645, opened 2023-06-30
Add click tracking to best bet boxes and expand all button

kewitham:


_assets/js/utils/gaUtil.js
```diff
@@ -19,6 +19,8 @@ function gtag() {
     ['.learn-link', 'click', sendGAClickEvent],
     ['.crt-sidenav-subnav-item', 'click', sendGAClickEvent],
     ['.info-box-link', 'click', sendGAClickEvent],
+    ['.best-bet', 'click', sendGAClickEvent],
```
okkays, line 22: any way to include info on which best bet they've clicked?

okkays (changes requested):
good idea!

_assets/js/utils/gaUtil.js
kewitham: Yes! The event will include the inner text of the link and the link address

okkays (approved):
very cool, can't wait to see some of this

PR #646, opened 2023-07-03
Add new best bet box type

kewitham:
Mock of new best bet box with links and no descriptions:
image

### DIFF
--- a/_assets/js/templates/search/graphicBestBetsSectionTemplate.js
+++ b/_assets/js/templates/search/graphicBestBetsSectionTemplate.js
@@ -1,0 +1,10 @@
+export default function bestBetsSectionTemplate(content) {
+    return `
+     <div class="padding-bottom-1 padding-top-4 usa-prose border-top-2px border-bottom-2px border-primary margin-top-5">
+     <p markdown="0" class="font-body-lg">Recommended by ADA.gov</p>
+        <div class="grid-row flex-wrap flex-justify">
+        ${content.join('')}
+        </div>
+     </div>
+     `
+ }
--- a/_assets/js/templates/search/graphicBestBetsTemplate.js
+++ b/_assets/js/templates/search/graphicBestBetsTemplate.js
@@ -1,12 +1,12 @@
 export default function graphicBestBetsTemplate(content) {
     const links = content.links.map(link => {
-        return `<a href="${link.url}" class="padding-right-2" aria-label="best bet box">${link.title}</a>`
+        return `<a href="${link.url}" class="padding-right-3 grid-col-6 padding-bottom-2" aria-label="best bet box">${link.title}</a>`
     });
     return `
-     <div class="padding-bottom-3 padding-top-2 usa-prose best-bet-item">
-       <p class="font-body-lg"><b class="title"><a href="${content.title_url}" class="best-bet" aria-label="best bet box">${content.title
+     <div class="padding-bottom-3 padding-top-2 usa-prose">
+       <p class="font-body-lg"><b class="title">${content.title
      .replace(/\uE000/g, '')
-     .replace(/\uE001/g, '')}</a></b></p>
+     .replace(/\uE001/g, '')}</b></p>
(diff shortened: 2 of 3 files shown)


okkays (approved):
LGTM

randyabramson:
Looks good to me @kewitham. Question - is the top line (Child Care) hyperlinked? If not, maybe we should take away the underline?

kewitham:
> Looks good to me @kewitham. Question - is the top line (Child Care) hyperlinked? If not, maybe we should take away the underline?

It is hyperlinked but we can also set it up to not be hyperlinked and have all the links below it.

nguyen-tina (commented):
Removing the larger 'Child Care' hyper link would also remove the link to the Child Care topics page. For that reason, I'm inclined to keep it! 

Visual question: Why is that text larger than the usual best bet link text size? Is it because it's a collection of links? Just want to clarify any new pattern that may be emerging here.

kewitham:
> 

@nguyen-tina we can keep the Child Care as just a plaintext title and add the child care topics page underneath if we want. The text for the links below the title link are the same size as the other best bet links but we have full control over the size so if you have a recommendation I can implement it.

nguyen-tina:
Got it – In that case, I would keep Child Care as a title (without the underline), and add the Child Care topics page as the first link of the set. Perhaps the title of that topics page could be, 'Child Care: Topic overview'

cc @vwilliamsn

vwilliamsn:
> Got it – In that case, I would keep Child Care as a title (without the underline), and add the Child Care topics page as the first link of the set. Perhaps the title of that topics page could be, 'Child Care: Topic overview'
> 
> cc @vwilliamsn

I like @nguyen-tina recommendations. Let's go with that.

kewitham:
> > Got it – In that case, I would keep Child Care as a title (without the underline), and add the Child Care topics page as the first link of the set. Perhaps the title of that topics page could be, 'Child Care: Topic overview'
> > cc @vwilliamsn
> 
> I like @nguyen-tina recommendations. Let's go with that.

updated!

vwilliamsn (approved):
Looks good thanks @kewitham!

kewitham:
> Looks good thanks @kewitham!

@vwilliamsn do we want to match the service animals best bet box to this one or keep it as is? Here is what it looks like now:
image
We probably should pick a consistent design across all best bet boxes.

vwilliamsn:
> > Looks good thanks @kewitham!
> 
> @vwilliamsn do we want to match the service animals best bet box to this one or keep it as is? Here is what it looks like now: image We probably should pick a consistent design across all best bet boxes.

Good point @kewitham!  I agree we need to have a consistent look and feel for the best bet boxes. With that in mind let's have the service animals best bet box format be our standard format for all of the best bet boxes.

kewitham:
@vwilliamsn sounds good! I forgot to note that the text best bet (the version we are using for service animals) only allows you to recommend two results whereas this graphic best bet version allows you to recommend as many results are you want. Wanted to flag in case that changes things.

vwilliamsn:
@kewitham thanks for pointing that out. I think providing the users with access to more content related to their search term, would be best.  So, I'm (now) inclined to go with the graphic option. Sorry for flip flopping on this. That also means the service animal best bet box would need to change and we can add a third option: [Frequently Asked Questions about Service Animals and the ADA](https://www.ada.gov/resources/service-animals-faqs/ ). Yay!  Happy to chat about this tomorrow at stand-up if you have any questions.

kewitham:
> @kewitham thanks for pointing that out. I think providing the users with access to more content related to their search term, would be best. So, I'm (now) inclined to go with the graphic option. Sorry for flip flopping on this. That also means the service animal best bet box would need to change and we can add a third option: [Frequently Asked Questions about Service Animals and the ADA](https://www.ada.gov/resources/service-animals-faqs/). Yay! Happy to chat about this tomorrow at stand-up if you have any questions.

sounds good! updated!

PR #648, opened 2023-07-13
Fix the notice banner, and remove the notice.

okkays:
(From a conversation with @c2nelson and @kewitham)

# What does this change?

- 🌎 The notice banner currently expects a notice to exist.
- ⛔ That's not great because there's not always something to alert about.
- ✅ This commit fixes the site so that the banner is hidden when there's no notice content

# Proof it works:

With nothing set or no posts:

Nothing

With notice_text and notice_link set:

Header + link

With only notice_link set:

Link only


c2nelson (approved):
thanks, @okkays.  looks good.

PR #650, opened 2023-07-18
Remove aria-label from learn more

okkays:
Make "learn more" links read correctly for screen readers.

- 🌎 These learn more links show up on the landing page. Right now, they have static aria-labels.
- ⛔ This means that they're all read as "learn more example link" to screen readers.
- ✅ This commit removes the aria labels, meaning the link text will be read instead (see example below)

| Before | After |
|--------|--------|
| example with reader showing bad aria-label | Example showing reader reading links without label |

### DIFF
--- a/_includes/landing/learn.html
+++ b/_includes/landing/learn.html
@@ -35,7 +35,7 @@
                   <ul class="padding-left-2">
                   {% for example in section.examples %}
                     <li class="padding-y-1">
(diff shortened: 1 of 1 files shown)


c2nelson (approved):
lgtm!  thanks.

vwilliamsn (approved):
Thanks @okkays!

PR #651, opened 2023-07-18
Analytics: Re-add link tracking to the learn more links

okkays:
- 🌎 The aria-label was used to provide an event name
- ⛔ Aria-label doesn't make sense for all links, and was removed.
- ✅ This commit re-adds an event name without an aria-label


kewitham (approved):
lgtm!

c2nelson (commented):
awesome!

PR #657, opened 2023-07-26
Adds external alert configuration.

okkays:
# What is this?

- 🌎 Our alerts link to notice posts
- ⛔ They _must_ link to notice posts - you can't link to somewhere external.
- ✅ This commit adds a way to add notices via _config.yml, instead

# Proof it works:

links


okkays:
@c2nelson - I've added an example alert in _config.yml here

Feel free to pull and change this branch with the new alert, or I can add the alert here if you give me the text.

PR #671, opened 2023-08-07
Subpart parsing

kewitham:
This PR adds logic to build larger subparts in the laws and regs to capture multiple sections

image


_assets/js/laws-regs-parser.js
```diff
@@ -1,51 +1,79 @@
...
         lastHeading = i;
-    });
+        length++
+        const section = createDiv('section');
+        section.innerHTML = sectionText.trim();
+        return section;
+    }).filter(section => !!section);
 
+    let currentSubpart = subparts[0];
     sections.map((section, i) => {
```
okkays, line 30: nit: prefer `forEach` when not using the return value of `map` (here and elsewhere)

okkays (approved):

PR #674, opened 2023-08-08
Netlify internal link bug fix

kewitham:
Currently there is a bug where Netlify is adding a \ character in front of links causing them to break. This PR adds JS to remove that character when the entry is saved on Netlify.
It also updates the Netlify config to correctly pull entries under notices.

### DIFF
--- a/_assets/js/netlify/makeHTMLFromBodyContent.js
+++ b/_assets/js/netlify/makeHTMLFromBodyContent.js
@@ -5,12 +5,10 @@ md.options['linkify'] = true;
 const parser = new DOMParser();
 function makeHTMLFromBodyContent(bodyContent, variables, imageData) {
   // Replace any <hr> tags with markdown so they render outside of the nearest <p> tag
-  let content = bodyContent
-    ? bodyContent
-      .replaceAll(/<hr>/g, "***")
+  let content = bodyContent?.replaceAll(/<hr>/g, "***")
       .replaceAll("{{'", '')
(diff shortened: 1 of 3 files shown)


c2nelson (approved):
thanks for fixing this bug!

_assets/js/netlify/makeHTMLFromBodyContent.js
```diff
@@ -5,7 +5,12 @@ md.options['linkify'] = true;
 const parser = new DOMParser();
 function makeHTMLFromBodyContent(bodyContent, variables, imageData) {
   // Replace any <hr> tags with markdown so they render outside of the nearest <p> tag
-  let content = bodyContent ? bodyContent.replaceAll(/<hr>/g, "***") : '';
+  let content = bodyContent
+    ? bodyContent
+      .replaceAll(/<hr>/g, "***")
+      .replaceAll("{{'", '')
+      .replaceAll("' | relative_url}}", '')
+    : '';
```
okkays, line 13: javascript has a "null coalescing" operator that might be useful here, as it's slightly more extensible than the ternary operator. If you do additional operations to bodyContent that might result in undefined / null, you can add more `?` rather than add more ternary clauses:

```suggestion
  let content = bodyContent?
      .replaceAll(/<hr>/g, "***")
      .replaceAll("{{'", '')
      .replaceAll("' | relative_url}}", '')
    || ''
```

okkays (approved):

PR #681, opened 2023-08-15
Carousel GA tracking class fix

kewitham:
This PR fixes an issue where the carousel-nav class needed for GA click tracking was being removed in the JS.


okkays (approved):
Whoops, good catch! Do we have a way to test out these events locally to make sure this doesn't happen?

kewitham:
> Whoops, good catch! Do we have a way to test out these events locally to make sure this doesn't happen?

Yes I just missed it when I was testing 😞

PR #682, opened 2023-08-16
Add feature flag and a/b testing

kewitham:
This PR enables feature flags to show features only to users who have logged into Netlify by adding a session variable. It also adds A/B testing capabilities by randomly assigning cookie A to 50% of users and cookie B to the other 50%.

### DIFF
--- a/_assets/js/feature-flag.js
+++ b/_assets/js/feature-flag.js
@@ -1,38 +1,39 @@
-export default function getCookies() {
+export default function setCookies() {
     const isNetlifyUser = sessionStorage.getItem('isNetlifyUser') ?? false;
-    const featureFlags = setFeatureFlagCookies(isNetlifyUser);
-    const variant = setVariant();
-
-    return {'variant': variant, 'feature_flags': featureFlags};
+    setFeatureFlagCookies(isNetlifyUser);
   }
 
-function setVariant() {
+function setPublicPercentVariant(name, publicPercentOn) {
   let variant;
   // Check if the user already has a variant assigned
   const userVariant = document.cookie
       .split("; ")
-      .find((row) => row.startsWith("ab_variant="))
+      .find((row) => row.startsWith(name + "="))
       ?.split("=")[1];
 
   if (userVariant) {
     variant = userVariant;
   } else {
     // Assign a random variant if the user doesn't have one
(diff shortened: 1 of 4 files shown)


_assets/js/feature-flag.js
```diff
@@ -0,0 +1,38 @@
...
+    // Assign a random variant if the user doesn't have one
+    variant = Math.random() < 0.5 ? 'A' : 'B';
+    // Set the variant in a cookie for future visits
+    document.cookie = 'ab_variant='+ variant;
+  }
+
+  return variant;
+}
+
+const FEATURE_FLAGS = [
```
okkays, line 29: WDYT about using this same list to assign different a/b test variants, as well?

I guess I'm thinking of "feature" as a singular object in the sense that you could:

1. define a "feature" (e.g., 'laws-and-regs')
2. turn that feature on or off for everyone (100% or 0%)
3. turn that feature on or off for a subset (i.e. 20%, 44%, etc; "a/b testing")
4. opt in/out of the feature as an admin ("feature flag")

okkays (changes requested):
nice work, very straightforward! One thought about the approach but the code is looking good!

_assets/js/feature-flag.js
```diff
@@ -0,0 +1,39 @@
...
+    {
+      name: 'laws-and-regs',
+      released: false,
+      publicPercentOn: 0,
+      optedIn: false,
+    }
```
okkays, line 30: 💅 so beautiful

_assets/js/feature-flag.js
```diff
@@ -0,0 +1,39 @@
...
+  const userVariant = document.cookie
+      .split("; ")
+      .find((row) => row.startsWith(name + "="))
+      ?.split("=")[1];
+
+  if (userVariant) {
+    variant = userVariant;
+  } else {
+    // Assign a random variant if the user doesn't have one
+    variant = Math.random() < publicPercentOn ? true : false;
```
okkays, line 18:
```suggestion
    variant = Math.random() < publicPercentOn;
```

okkays (approved):
Nice! A few nits but this LGTM, and is so powerful for only 55 LOC.

PR #689, opened 2023-09-06
Add checkbox to admin page for feature flags

kewitham:
This PR updates the logic for the feature flag to use a checkbox selection rather than a window event listener. Checking the box will allowlist the Netlify user on the feature and unchecking it will blocklist them. Once the session expires the user will blocklisted until they check the box again.

image


okkays (approved):
Nice! It might be good to add some docs for future devs on how to create/use these

PR #694, opened 2023-09-20
Touchpoints yes/no poll

kewitham:
This PR adds the "Was this page helpful" touchpoints question for pages with and without a sidenav.

image

image

### DIFF
--- a/_assets/js/touchpoints-page-helpful.js
+++ b/_assets/js/touchpoints-page-helpful.js
@@ -7,5 +7,14 @@ export default function tryToSetPageAnswer() {
       return;
     }
 
+    const answer1 = document.querySelector('#touchpoints-yes-no-form #answer_01');
+    const submitButton = document.querySelector('#touchpoints-yes-no-form .submit_form_button');
+
+    Array.from(answer1.getElementsByTagName('input')).forEach(input => {
+      input.addEventListener('change', () => {
+        submitButton.click();
+      })
+    });
+
     pageAnswer.value = window.location.href;
   }
--- a/_assets/js/touchpoints-page-helpful.js
+++ b/_assets/js/touchpoints-page-helpful.js
@@ -1,0 +1,11 @@
+export default function tryToSetPageAnswer() {
+    if (!document.querySelector('#touchpoints-yes-no-form')) return;
+
+    const pageAnswer = document.querySelector('#touchpoints-yes-no-form #answer_02');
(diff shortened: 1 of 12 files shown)


_layouts/default.html
```diff
@@ -8,6 +8,9 @@
   <main id="main-content" class="margin-top-0 margin-bottom-0">
     {% include notice.html %}
     {{ content }}
+    {% if page.page-helpful != false %}
```
okkays, line 11:
```suggestion
    {% if page.page-helpful %}
```

_layouts/page.html
```diff
@@ -23,6 +23,9 @@ <h4 class="usa-accordion__heading desktop:display-none">
       {% include sidenav.html collection=page.subnav %}
       </div>
       {% endif %}
+      {% if page.page-helpful != false and page.permalink contains '/law-and-regs/' %}
```
okkays, line 26:
```suggestion
      {% if page.page-helpful and page.permalink contains '/law-and-regs/' %}
```

okkays (changes requested):
Do you think it'd be worth it to hide "Submit" and add a click handler to the "Yes / No" which automatically clicks the "Submit"? (That way we could hide the submit button and it'd just have the "Yes/No"

_layouts/default.html
kewitham: The poll won't show up if I do this because the attribute is only set if it's false.

vwilliamsn (approved):
LGTM! Thanks @kewitham!

sanchez-amanda (approved):
Looks good!!!

_assets/js/touchpoints-page-helpful.js
```diff
@@ -0,0 +1,20 @@
+export default function tryToSetPageAnswer() {
+    if (!document.querySelector('#touchpoints-yes-no-form')) return;
+
+    const pageAnswer = document.querySelector('#touchpoints-yes-no-form #answer_02');
+    if (!pageAnswer) {
+      setTimeout(tryToSetPageAnswer, 500);
+      return;
+    }
+
+    const answer1 = document.querySelector('#touchpoints-yes-no-form #answer_01');
```
okkays, line 10: supernit - wdyt about naming this "yesNoButtons" or something instead of answer1 to give a bit more context to future readers

okkays (approved):

PR #704, opened 2023-10-10
Update copy and print for law and regs

kewitham:
This PR fixes two issues with the law and regs copy/print functionality:
1. Resolves a bug where the user was directed to a 404 page after printing
2. Opens accordions when copying and printing so all content within the accordions is captured.

### DIFF
--- a/_assets/js/laws-regs-parser.js
+++ b/_assets/js/laws-regs-parser.js
@@ -81,7 +81,7 @@ function buildBtns(i, divType) {
     const btnDiv = createDiv('btn-group display-flex flex-row flex-justify maxw-card');
     const btnTypes = ['Share', 'Copy', 'Print'];
     btnTypes.forEach(btnType => {
-        const btn = document.createElement('div');
+        const btn = document.createElement('a');
         btn.className = btnType + '-btn text-no-underline section-btn';
         const gaEventName = 'data-ga-event-name="' + btnType + ' ' + divType + ' ' + i + '"';
         btn.setAttribute('aria-label', btnType);
--- a/_assets/js/laws-regs-parser.js
+++ b/_assets/js/laws-regs-parser.js
@@ -157,13 +157,13 @@ function unHighlightText(e, divType) {
 }
 
 function openAccordions(section) {
-    Array.from(section.getElementsByTagName('details')).forEach(accordion => {
+    section.querySelectorAll('details').forEach(accordion => {
         accordion.setAttribute('open', true);
     });
 }
 
 function closeAccordions(section) {
-    Array.from(section.getElementsByTagName('details')).forEach(accordion => {
+    section.querySelectorAll('details').forEach(accordion => {
         accordion.removeAttribute('open');
     });
 }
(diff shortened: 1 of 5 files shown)


vwilliamsn:
@kewitham - thanks for working on this!  Not sure why, but the preview in federalist link isn't working properly.

_assets/js/laws-regs-parser.js
```diff
@@ -1,4 +1,8 @@
 export default function parseLawsAndRegs (mainContent) {
+    const location = window.location.href;
+    if (location.includes('SVGAnimatedString')) {
+        history.back();
```
okkays, line 4: does this block until after the results of going back are definitely done? (navigation like `window.location.reload()`, etc, doesn't always prevent the JS from continuing to run)

If not, maybe add a block after such as:

```
if (the SVGAnimatedString is still visible / preventing what we want) {
  setTimeout(() => parseLawsAndRegs(mainContent), 500);
  return;
}
```

_assets/js/laws-regs-parser.js
```diff
@@ -83,15 +87,15 @@ function buildBtns(i, divType) {
...
+            btn.innerHTML = '<span class="copied text-no-underline" style="display:none;">Copied text</span><svg title="Copy text" ' + gaEventName + ' class="usa-icon copy-icon usa-tooltip" data-position="bottom" focusable="false" role="img"><title>Copy text</title><use xlink:href="/assets/img/sprite.svg#content_copy"></use></svg>';
             btn.addEventListener('click', (e) => { copyText(e, divType) });
             btn.addEventListener('mouseover', (e) => { highlightText(e, divType) });
             btn.addEventListener('mouseleave', (e) => { unHighlightText(e, divType) });
         } else if (btnType === 'Print') {
-            btn.innerHTML = '<svg title="print" ' + gaEventName + ' class="usa-icon usa-tooltip" data-position="bottom" focusable="false" role="img"><title>print</title><use xlink:href="/assets/img/sprite.svg#print"></use></svg>';
+            btn.innerHTML = '<svg title="Print" ' + gaEventName + ' class="usa-icon usa-tooltip" data-position="bottom" focusable="false" role="img"><title>Print</title><use xlink:href="/assets/img/sprite.svg#print"></use></svg>';
```
okkays, line 98: maybe this would be better as a template string, so it could be multiline / help with interpolation (since html whitespace shouldn't matter here)?
```suggestion
            btn.innerHTML = `
              <svg
                title="Print"
                ${gaEventName}
                class="usa-icon usa-tooltip"
                data-position="bottom"
                focusable="false" role="img">
                  <title>Print</title>
                  <use xlink:href="/assets/img/sprite.svg#print">
                  </use>
                </svg>
              `;
```

_assets/js/laws-regs-parser.js
```diff
@@ -111,9 +115,22 @@ function unHighlightText(e, divType) {
     section.classList.remove('highlight');
 }
 
+function openAccordions(section) {
+    Array.from(section.getElementsByTagName('details')).forEach(accordion => {
```
okkays, line 119: supernit: a shorter way to write this, as querySelectorAll has forEach built in:

```suggestion
    section.querySelectorAll('details').forEach(accordion => {
```

okkays (changes requested):

okkays (approved):

PR #721, opened 2023-11-16
Switch to Ruby 3

okkays:
This switches us to ruby 3.2.2, mostly following [this cloud.gov guidance](https://cloud.gov/2023/05/24/cloud-gov-pages-jekyll-ruby-upgrade/), but with some special consideration for webpack.

This involves removing things that don't support ruby 3.2.2, especially the jekyll-assets plugin. Switching up the way we do assets is a _big change_, because we have to go through everywhere in the site that an image is used and point it to a new url.

So, this PR does that! We've:

1. Upgraded to Ruby 3, and upgraded our Gemfile to pull in any support for the new Ruby.
2. Moved our assets to the _assets folder.
3. Updated references to assets in html, md, css, and javascript (**this is a majority of the changes**)
4. Modified our webpack configuration to support moving the select images and fonts we use from USWDS
5. Modified webpack further to support the postprocessing (i.e., automatically adding css rules to support old browsers / new css features)
6. Tried everything out, debugged, etc (including as much as I could in netlify)

This PR does **not**:

- Keep support for `{% asset %}`. It would have been non-trivial to support it because of the asset pipeline required to parse asset paths under the hood.
- Keep support for asset hashing. We may want to look at ways to re-add this one day, but it's not a trivial ask given our combination of jekyll + webpack.


okkays:
@c2nelson @kewitham we can discuss the extent of this one in a pairing session 😅 

It's a lot of files changed but the new way of including assets is still pretty straightforward

c2nelson:
Thank you, @okkays, for all of your work on this and detailing it for us in our sync.  Super helpful.  Great job!

kewitham (approved):
works as expected, lgtm! 🎉

PR #726, opened 2023-12-01
Enforcement ga click tracking

kewitham:
Add click tracking to enforcement page
Note: because of some issues resolving merge conflicts with the new ruby version, some formatting changes made their way into this PR (but that's all they are)


okkays (approved):
Thanks for checking on the formatting changes!

PR #727, opened 2023-12-01
Add custom field for faceted search

kewitham:
This PR adds a custom meta search field for search.gov to scrape so we will be able to display meta descriptions as the search result descriptions instead of snippets when they are available.


c2nelson:
@kewitham In this preview the news item is back.  Am I looking at an older view?  I removed it a while ago.

kewitham:
> @kewitham In this preview the news item is back. Am I looking at an older view? I removed it a while ago.

@c2nelson on the home page? I'm not seeing that. Maybe you checked the federalist link before my updates were pushed.

assets/js/netlify/makeHTMLFromBodyContent.js
```diff
@@ -8,6 +8,8 @@ function makeHTMLFromBodyContent(bodyContent, variables, imageData) {
   let content = bodyContent?.replaceAll(/<hr>/g, "***")
       .replaceAll("{{'", '')
       .replaceAll("' | relative_url}}", '')
+      .replaceAll("'| relative_url}}", '')
+      .replaceAll("'|relative_url}}", '')
```
okkays, line 12: You could use a regex for this, too - that would also account for the ability to have spaces before the `}}`, and to have an arbitrary number of spaces:

https://regex101.com/r/EjI2lY/2

```suggestion
      .replaceAll(/\s*\|\s*relative_url\s*}}/, '')
```

okkays (approved):
LGTM, one comment

PR #733, opened 2024-01-05
Enable Javascript Linter

okkays:
- 🌎 Our javascript has no enforced formatting
- ⛔ This is problematic because it means automatic linters will pick up changes unrelated to the current code change.
- ✅ This commit turns on prettier linting using github super linter.

To automatically fix javascript code to match prettier styles, there's an npm shortcut:

```sh 
npm run lint:write  # To auto-format code
npm run lint:check  # To see if code meets formatting guidelines
```


kewitham (approved):
lgtm!

PR #743, opened 2024-01-22
Update field/styles for faceted search

kewitham:


assets/js/templates/search/searchResultsTemplate.js
```diff
@@ -1,14 +1,23 @@
+function getDescription(content) {
+  if (content.searchgov_custom1) {
+    const custom_description = content.searchgov_custom1[0].replace(',', ', ');
+    return custom_description.charAt(0).toUpperCase() + custom_description.slice(1);
+  } else {
+    return content.snippet
+      .replace(/\uE000/g, '<span class="bg-yellow">')
+      .replace(/\uE001/g, '</span>');
+  }
+}
```
okkays, line 10: unnecessary else after return:

```suggestion
  if (content.searchgov_custom1) {
    const custom_description = content.searchgov_custom1[0].replace(',', ', ');
    return custom_description.charAt(0).toUpperCase() + custom_description.slice(1);
  }
  
  return content.snippet
    .replace(/\uE000/g, '<span class="bg-yellow">')
    .replace(/\uE001/g, '</span>');
}
```

okkays (approved):
One quick style nit, otherwise LGTM!

PR #755, opened 2024-02-09
Fix TTY #

kewitham:


vwilliamsn (approved):
You're the best, thanks @kewitham!

PR #766, opened 2024-03-05
T2 Primer typo fix

c2nelson:
redundant text removed.


vwilliamsn (approved):
@c2nelson - thanks for getting this done so quickly!

PR #773, opened 2024-03-22
Add title ii and iii to manual_redirects

okkays:
- 🌎 We recently changed the URLs for title ii and title iii
- ⛔ This causes problems for folks with bookmarked pages
- ✅ This commit adds a manual redirect from the old urls to the new ones.

Note that this will still result in 404's in the logs. They are redirected following the 404 via javascript.

### DIFF
--- a/_data/manual_redirects.json
+++ b/_data/manual_redirects.json
@@ -1,4 +1,6 @@
 {
   "/index.html": null,
-  "/404.html": null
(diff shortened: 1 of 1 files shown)


kewitham (approved):

vwilliamsn (approved):
LGTM, thanks @okkays!

PR #777, opened 2024-03-29
Button fix

kewitham:
This PR takes another pass at fixing the back to top button on mobile. The original attempt moved the button to be on screen but this had side effects for the feedback button on desktop. This PR makes it so the content is only the width of the viewport (by wrapping tables in scrollable divs and wrapping long links) so the button doesn't end up too far to the right.


okkays (approved):
LGTM!

PR #778, opened 2024-04-02
Direction 2: Law & regs tooling

kewitham:
This PR adds in page search functionality to the law and regs side nav.
Future PRs will build the mobile view and the sticky view.

Image

### DIFF
--- a/assets/js/laws-regs-parser.js
+++ b/assets/js/laws-regs-parser.js
@@ -103,9 +103,10 @@ function buildBtns(i, divType) {
     btn.className = btnType.toLowerCase() + '-btn text-no-underline section-btn';
     const gaEventName = 'data-ga-event-name="' + btnType + ' ' + divType + ' ' + i + '"';
     btn.setAttribute('aria-label', btnType);
+    btn.href = '#';
     if (btnType === 'Share') {
       btn.innerHTML = `
-                <p class="copied-link text-no-underline margin-0" style="display:none;">Copied link</p>
+                <span class="copied-link text-no-underline" style="display:none;">Copied link</span>
                 <svg title="Copy link"
                     ${gaEventName}
                     class="usa-icon share-icon usa-tooltip"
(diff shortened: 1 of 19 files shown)


kewitham:
> (a11y tests are failing)

it looks like this is related to the copy/print/share buttons (seems to fail intermittently for these)

assets/js/laws-regs-parser.js
```diff
@@ -242,3 +235,157 @@ function shareLink(e, divType) {
...
+    if (newCount <= 0) {
+      prevBtn.classList.add('disabled');
+    }
+    if (newCount < totalCount) {
+      nextBtn.classList.remove('disabled');
+    }
+  }
+}
+
+function search() {
```
okkays, line 299: this function is pretty large

I feel like there are a couple of different things going on here (highlighting, dealing with buttons)

Splitting those off might make it a bit more readable

assets/js/laws-regs-parser.js
```diff
@@ -242,3 +235,157 @@ function shareLink(e, divType) {
...
+      section.classList.remove('searched');
+      const detailEls = section.querySelectorAll('details');
+      detailEls.forEach((detailEl) => detailEl.removeAttribute('open'));
+    }
+  });
+  const results = document.querySelectorAll('.search-term');
+  if (results.length > 0) {
+    results.forEach((result, i) => {
+      const count = i + 1;
+      result.id = 'result' + count;
```
okkays, line 330: I worry about calling this `result` - there's a good chance someone adds some other `result` in the future that conflicts.

Does this have to be an ID, or can you make it a class and use

```
someContainer.querySelector(`.result-${i}`)
```

to find it instead?

okkays (changes requested):

assets/js/laws-regs-parser.js
kewitham: I used id so I can use location.hash to take the user to the search term

okkays (approved):
LGTM!

vwilliamsn (approved):
LGTM, thanks @kewitham!

PR #795, opened 2024-04-18
Add documentation for content grid

kewitham:
This PR updates the readme to include information about the content grid generated with each PR.


okkays (approved):
Thanks for adding this!

This works if it's documentation for us - if we want non-technical folks to find it we might need a GIF or something.

PR #782, opened 2024-04-05
Law & regs tooling - mobile

kewitham:
This PR adds the mobile functionality for the law and regs tooling:
Image

Image

Image

PR #807, opened 2024-06-20
Law and regs tooling updates

kewitham:
This PR moves all the law and regs tooling updates (#778 #782) into one PR and rebases since the main branch has changed a lot since their initial creation.


okkays (approved):
LGTM! Noting that the copy, print, highlight features have missing icons on the pages preview, but are still present and functional.

PR #808, opened 2024-06-24
Integrated text

c2nelson:
This PR adds the new web rule text to the existing T2 reg page and replaces the pdfs.


vwilliamsn (approved):
Great job @c2nelson!

PR #811, opened 2024-07-15
Fix+maintenance

c2nelson:
Updated UA content and changed date of T2 Web fact sheet.


vwilliamsn (approved):
LGTM, thanks @c2nelson!

PR #819, opened 2024-07-26
mde press release link added to banner

c2nelson:
The AG signed the MDE rule.  Replaced the anniversary signing link with the press release re the rule.


vwilliamsn (approved):
Thanks @c2nelson!

PR #820, opened 2024-07-26
removed news items

c2nelson:
Removed items from New on ADA.gov feature because they were are no longer new.


vwilliamsn (approved):
Thank you!

PR #850, opened 2024-11-22
[FAQ 4] Add content for help resources

okkays:
- 🌎 This content goes in the new help/FAQ page
- ✅ This commit adds a list of links between topics and relevant standards sections

Screenshot 2024-11-22 at 3 16 27 PM

Screenshot 2024-11-22 at 3 16 41 PM

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -71,6 +71,17 @@ primary_navigation:
     url:
       en: "/resources/"
       es: "/resources/"
+    children:
+    - name:
+        en: Guidance & Resource Materials
+        en-short: Resources
+        es: Guidance & Resource Materials
+      url: "/resources/"
+    - name:
+        en: Help
+        en-short: Help
+        es: Help
+      url: "/help/"
   - name:
       en: Law, Regulations & Standards
       en-short: Law & Regulations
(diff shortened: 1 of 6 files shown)


okkays:
(Pending feedback from https://github.com/usdoj-crt/beta.ADA.gov-Project-Management/issues/1036#issuecomment-2494651934)

_config.yml
```diff
@@ -71,6 +71,17 @@ primary_navigation:
...
+    - name:
+        en: Guidance & Resource Materials
+        en-short: Resources
+        es: Guidance & Resource Materials
+      url: "/resources/"
+    - name:
+        en: Help
+        en-short: Help
+        es: Help
+      url: "/help/"
```
Rwolfe-Nava, line 84: Just checking: are `en` and `es` here supposed to be the same? Is there some other system that is handling the translation for us?

Rwolfe-Nava (approved):
Just a small question regarding translation, but otherwise this looks good to me

_config.yml
okkays: This is where we'd put the translations, but that would be separate work to procure those - even for a single word like this - so we'll default to en for now.

